PR #99, opened 2020-08-06
En and fr translations

Pollux3737:
Here is a translation to French and English of the guide document (dating of 2020/08/06)


SebastianMestre:
Hey! Thank you for taking the time to do this, I very much appreciate it.

PR #101, opened 2020-08-07
Typechecker integration

SebastianMestre:
Oh boy did debugging this take some work.

I finally started putting and propagating types through the syntax tree. It's pretty buggy and incomplete, so it fails the tests quite catastrophically. I did get it to typecheck this program, though:

```rust
f := fn(x) => x;

__invoke := fn(){
	return f(25);
};
```

UPDATE: this typechecks too, now:

```rust
f := fn(x, y) {
	return x + y;
};

__invoke := fn(){
	x := 5;
	return f(x, 25);
};
```

UPDATE(2020-08-09):
this now gets rejected at the type checker
```rust
__invoke := fn(){
  x := "A";
  return x + 5;
}
```
while this finishes successfully
```rust
__invoke := fn(){
  x := 10;
  return x + 5;
}
```

UPDATE: this now gets rejected by the type checker
```rust
f := fn(x, y){
  return x + y;
};
__invoke := (){
  a := "a";
  b := 5;
  return f(a, b);
}
```

### DIFF
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -7,21 +7,20 @@ namespace Frontend {
 CompileTimeEnvironment::CompileTimeEnvironment() {
 	// TODO: put this in a better place
 	// HACK: this is an ugly hack. bear with me...
-	static TypedAST::Declaration dummy;
-	declare("size", &dummy);
-	declare("print", &dummy);
-	declare("array_append", &dummy);
-	declare("array_extend", &dummy);
-	declare("array_join", &dummy);
+	declare_builtin("size");
+	declare_builtin("print");
+	declare_builtin("array_append");
+	declare_builtin("array_extend");
+	declare_builtin("array_join");
 
-	declare("+", &dummy);
-	declare("-", &dummy);
-	declare("*", &dummy);
-	declare("/", &dummy);
-	declare("<", &dummy);
-	declare("=", &dummy);
-	declare("==", &dummy);
-	declare(".", &dummy);
+	declare_builtin("+");
+	declare_builtin("-");
+	declare_builtin("*");
+	declare_builtin("/");
+	declare_builtin("<");
+	declare_builtin("=");
+	declare_builtin("==");
+	declare_builtin(".");
 };
 
 Scope& CompileTimeEnvironment::current_scope() {
@@ -30,6 +29,11 @@ Scope& CompileTimeEnvironment::current_scope() {
 
 void CompileTimeEnvironment::declare(std::string const& name, TypedAST::Declaration* decl) {
 	current_scope().m_vars[name] = decl;
+}
+
+void CompileTimeEnvironment::declare_builtin(std::string const& name){
+	static TypedAST::Declaration dummy;
+	declare(name, &dummy);
 }
 
 TypedAST::Declaration* CompileTimeEnvironment::access(std::string const& name) {
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -15,7 +15,6 @@ Binding::Binding(TypedAST::FunctionLiteral* func, int arg_index)
     , m_arg_index { arg_index } {}
 
 TypedAST::Declaration* Binding::get_decl() {
-	if(m_type != BindingType::Declaration){ *(int*)0 = 0; }
 	assert(m_type == BindingType::Declaration);
 	return m_decl;
 }
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -2,7 +2,33 @@
 
 #include "typed_ast.hpp"
 
+#include <cassert>
+
 namespace Frontend {
+
+Binding::Binding(TypedAST::Declaration* decl)
+    : m_type { BindingType::Declaration }, m_decl { decl } {}
+
+Binding::Binding(TypedAST::FunctionLiteral* func, int arg_index)
+    : m_type { BindingType::Argument }
+    , m_func { func }
+    , m_arg_index { arg_index } {}
+
+TypedAST::Declaration* Binding::get_decl() {
+	if(m_type != BindingType::Declaration){ *(int*)0 = 0; }
+	assert(m_type == BindingType::Declaration);
+	return m_decl;
+}
+
+TypedAST::FunctionArgument& Binding::get_arg() {
+	assert(m_type == BindingType::Argument);
+	return m_func->m_args[m_arg_index];
+}
+
+TypedAST::FunctionLiteral* Binding::get_func() {
+	assert(m_type == BindingType::Argument);
+	return m_func;
+}
 
 CompileTimeEnvironment::CompileTimeEnvironment() {
 	// TODO: put this in a better place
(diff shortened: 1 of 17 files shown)


src/match_identifiers.cpp
```diff
@@ -23,6 +23,10 @@ void match_identifiers(TypedAST::Declaration* ast, Frontend::CompileTimeEnvironm
 	env.declare(ast->identifier_text(), ast);
 	if (ast->m_value) {
 		match_identifiers(ast->m_value.get(), env);
+		// TODO: should this be a Var instead of just an id copy?
+		ast->m_value_type = ast->m_value->m_value_type;
```
lushoBarlett: Sí. Si hacés una copia, y después el type id del rhs lo pisás al hacer el union find, después el valor del decl queda desactualizado. Sería mejor el Var así dice "tengo el mismo tipo que el rhs"

src/typesystem.hpp
```diff
@@ -59,10 +60,13 @@ struct TypeSystemCore {
 
 	MonoId new_var();
 	MonoId new_term(TypeFunctionId type_function, std::vector<int> args);
```
lushoBarlett, line 62: En esta linea, el int no debería ser reemplazado por MonoId? (Sé que no es de este pr, solamente lo menciono)

src/typesystem.cpp
```diff
@@ -28,8 +28,44 @@ MonoId TypeSystemCore::new_term(TypeFunctionId type_function, std::vector<int> a
...
+	} else {
+		TermId term = data.data_id;
+		for (MonoId arg : term_data[term].arguments)
+			gather_free_vars(arg, free_vars);
+	}
+}
+
+// qualifies all free variables in the given monotype
+// TODO(Mestre): I dont think this is right, we are supposed to only qualify
+// 'unbound' variables... whatever than means.
```
lushoBarlett, line 55: Parafraseando lo que dice [acá](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system#Free_type_variables):

> Cualquier ocurrencia de \alpha_i en \tau está _bound_. A los _unbound_ type vars se les dice _free_.

Creo que es lo mismo. Además me parece que entre `gather_free_vars` y `new_poly` está escondida la `gen_rule`. Quizás estoy hablando sin saber, pero me suena a eso

lushoBarlett: Leyendo más detenidamente más abajo al costado, debo estar súper equivocado sobre la `gen_rule`, y no estoy más seguro sobre lo que dije de _free_, pero tengo más confianza en ese statement. HM no es fácil, pero por lo menos ahora parece que la página de Wikipedia la entiendo mejor.

src/typesystem.cpp
```diff
@@ -44,7 +53,7 @@ void TypeSystemCore::gather_free_vars(MonoId mono, std::unordered_set<VarId>& fr
 // qualifies all free variables in the given monotype
 // TODO(Mestre): I dont think this is right, we are supposed to only qualify
 // 'unbound' variables... whatever than means.
-PolyId TypeSystemCore::new_poly(MonoId mono) {
+PolyId TypeSystemCore::generalize(MonoId mono) {
```
lushoBarlett: Ah esto me refería con la `gen_rule`

src/match_identifiers.cpp
```diff
@@ -45,7 +45,7 @@ void match_identifiers(TypedAST::Identifier* ast, Frontend::CompileTimeEnvironme
 	}
 
 	// TODO: should this be a Var instead of just an id copy?
-	ast->m_value_type = ast->m_declaration->m_value_type;
+	ast->m_value_type = env.m_typechecker.m_core.inst_fresh(declaration->m_decl_type);
```
lushoBarlett, line 48: Acceder al core desde afuera no es ideal, quizás habría que hacer una interfaz para esto

lushoBarlett:
Bueno entendí más de lo que creí posible. Está tomando bastante forma esto.

src/match_identifiers.cpp
SebastianMestre: This is basically the [Let] deduction rule, just lazily thrown together

src/match_identifiers.cpp
SebastianMestre: I don't think this is how that works

src/typesystem.hpp
SebastianMestre: yes

src/compile_time_environment.cpp
```diff
@@ -13,14 +13,29 @@ CompileTimeEnvironment::CompileTimeEnvironment() {
...
-	declare_builtin(".");
+	// TODO: refactor, figure out a nice way to build types
+	auto var_mono_id = m_typechecker.new_var();
+	auto var_id = m_typechecker.m_core.mono_data[var_mono_id].data_id;
+
+	// TODO: i use the same mono thrice... does this make sense?
+	auto term_mono_id = m_typechecker.m_core.new_term(
+		TypeChecker::BuiltinType::Function,
+		{var_mono_id, var_mono_id, var_mono_id},
+		"[builtin] (a, a) -> a");
```
lushoBarlett, line 50: I would think so... This represents a binary operation on a set that is represented by the type, and the type is not known so we use a var... Shouldn't it be a poly? And later on do an Inst for each use? Else the deduction will fail if it's used for different types... Right?

lushoBarlett: Ah but it's a term that takes 3 vars as arguments, I think this is correct. We'll see

SebastianMestre: a poly is created from this mono on the very next line

src/match_identifiers.cpp
```diff
@@ -120,16 +153,15 @@ void match_identifiers(TypedAST::DeclarationList* ast, Frontend::CompileTimeEnvi
...
-			    = env.m_typechecker.m_core.generalize(d->m_value->m_value_type);
-		} else {
-			// TODO SPEED: dont create a variable and then generalize it, create a
-			// generalized variable directly
-			d->m_decl_type
-			    = env.m_typechecker.m_core.generalize(env.m_typechecker.new_var());
-		}
+
+		MonoId mono = d->m_value ? d->m_value->m_value_type
+			: env.m_typechecker.new_var();
```
lushoBarlett, line 186: You changed the ternary operator writing style here hahahaha. I personally like the ? and : being either on the same line or lined up (the latter being your other use case)

src/compile_time_environment.cpp
```diff
@@ -2,8 +2,34 @@
...
+
+Binding::Binding(TypedAST::FunctionLiteral* func, int arg_index)
+    : m_type { BindingType::Argument }
+    , m_func { func }
+    , m_arg_index { arg_index } {}
+
+TypedAST::Declaration* Binding::get_decl() {
+	if(m_type != BindingType::Declaration){ *(int*)0 = 0; }
+	assert(m_type == BindingType::Declaration);
+	return m_decl;
```
lushoBarlett, line 20: What is this???

SebastianMestre: Triggering the address sanitizer to get a stack trace

src/match_identifiers.cpp
SebastianMestre: I like putting things in different lines but this is how it's set up in the .clang-format file

src/compile_time_environment.cpp
lushoBarlett: Ah, now I get it

src/match_identifiers.cpp
lushoBarlett: Then the other use case did not get formatted

src/compile_time_environment.cpp
lushoBarlett: Clever, please comment `// address sanitizer` or something next time

SebastianMestre: It's debug code, it won't be there one the final pr

SebastianMestre:
> the changes are getting spread out ...

yeah, it's hard to keep it contained: there are a lot of changes required to make things work, and the tests won't pass until it's all (or almost all) done.

at the very least, I will not be able to squeeze recursion (it would involve building the declaration graph, splitting it into strongly-connected components, then sorting it topologically to finally process the declarations in groups), or non-crashy errors (tons of work, no idea how to approach it) into this pr

so, it's kind of a given that some of the tests will have to be temporarily commented out for this pr

PR #103, opened 2020-08-14
[WIP] Implement Kosaraju's algorithm

SebastianMestre:
I have implemented [Kosaraju's algorithm](https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm) for decomposition into strongly connected components. This will be used to determine how declarations should be grouped into rec blocks during type checking.

I was originally going to implement the [rec] rule from Hindley-Milner but, would it be better to keep this pr short by only implementing this algorithm, then do other pieces of work in different prs?

UPDATE:

I had the idea of building up the DAG of strongly connected components using Kosaraju's Algorithm, then finding a topological sort of it, to apply the rec rule in that order.

However, I recently learned that [Tarjan's Algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) actually gives the strongly connected components topologically sorted without any extra processing. Furthermore, Wikipedia claims that Tarjan's algorithm is more efficient in practice.

I hesitate to actually implement Tarjan's Algorithm because it is a bit more complicated, but it might be worth it.


SebastianMestre:
I actually learned how to implement Tarjan's algorithm, and it is actually not only more efficient, but also smaller, code-wise. I am scrapping this.

PR #105, opened 2020-08-20
Differentiate Integers from Numbers

SebastianMestre:
It makes a lot of sense to separate Integer literals from other number literals. This makes it so that we can, for example, deduce the type of NumberLiteral to Float, which is nice.


SebastianMestre:
> Good stuff. On an unrelated note, we might want to stop using [WIP] in pr names and start using drafts, I think. The GitHub recommendation for converting to draft is "still in progress?", that makes me think any WIP is a draft by definition. What do you think?

I've been doing both, lately.

PR #106, opened 2020-08-28
Put the interpreter in namespace Interpreter

SebastianMestre:
I am trying to sort out the namespaces, I started by putting all of the interpreter into the all-new `namespace Interpreter`

I was thinking about cleaning up the whole `AST-TypedAST-Frontend-TypeChecker` mess as well. I don't think we should have that much stuff in the top-level namespace.

Ideally, Jasper should be embeddable in other programs, so it will all have to be put into a `Jasper` namespace as well, eventually.

### DIFF
--- a/src/interpreter/environment.cpp
+++ b/src/interpreter/environment.cpp
@@ -4,7 +4,7 @@
 
 #include "garbage_collector.hpp"
 
-namespace Type {
+namespace Interpreter {
 
 void Scope::declare(const Identifier& i, Reference* v) {
 	// TODO: check name collision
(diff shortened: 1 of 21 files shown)


src/interpreter/environment.cpp
```diff
@@ -4,7 +4,7 @@
 
 #include "garbage_collector.hpp"
 
-namespace Type {
+namespace Interpreter {
```
lushoBarlett, line 7: This here is some legacy mess, I regret putting `Environment` in the `Type` namespace every day

PR #108, opened 2020-08-30
Tests decouple

lushoBarlett:
I'm trying to use polymorphism to generalize testing, but I can't seem to understand how to store pointers when given r-values? Maybe I should use `new` and even `unique_ptr` directly in main. This is new territory for me so I'm asking for help.

I tried storing r-values but I think that is not allowed, so the only option is a pointer (or `unique_ptr`).

### DIFF
--- a/src/test/main.cpp
+++ b/src/test/main.cpp
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <cassert>
+#include <memory>
 
 #include "../interpreter/environment_fwd.hpp"
 #include "../interpreter/execute.hpp"
@@ -7,13 +8,13 @@
 #include "tester.hpp"
 
 int main() {
-	using Test::TestInterpreter;
+	using Test::InterpreterTestSet;
 
 	Test::Tester tests;
 
 	/*
 	tests.add_test(
-		TestInterpreter{R"(
+		std::make_unique<InterpreterTestSet>(InterpreterTestSet{R"(
 			x : dec = 1.4;
 			y : int = 3;
 			z := fn () {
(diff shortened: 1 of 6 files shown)


lushoBarlett:
Maybe I should store `const TestSet&` since r-values are castable to that? But will it work for polymorphism?

lushoBarlett:
Should I do something like this?
```c++
Tester::Tester(TestSet&& ts) : m_runners{ make_unique<TestSet>(ts) } {}

Tester::Tester(std::vector<TestSet&&> tss) {
    for (auto&& ts : tss)
        Tester(ts);
}
```

SebastianMestre:
pass by value, move as needed to get rid of copies

lushoBarlett:
Thanks for the help!

src/test/main.cpp
```diff
@@ -76,12 +77,12 @@ int main() {
...
 			eval_expression("__invoke()", env);
 			return exit_status_type::Ok;
-		}}
+		}})
 	);
 	*/
 
 	tests.add_test(
-		TestSet{R"(
+		std::make_unique<InterpreterTestSet>(InterpreterTestSet{R"(
```
SebastianMestre, line 85: Make_unique takes the same arguments as the constructor of the type.

There is no need to do `make_unique<T>(T{ ... })`, because `make_unique<T>(...)` suffices

src/test/tester.cpp
```diff
@@ -3,57 +3,60 @@
...
 
-#include "../interpreter/exit_status_type.hpp"
+#include "test_status.hpp"
 
 namespace Test {
 
-Tester::Tester(TestSet ts) : m_test_sets({ std::move(ts) }) {};
-Tester::Tester(std::vector<TestSet> tss) : m_test_sets(std::move(tss)) {};
+template<typename T>
+using Own = std::unique_ptr<T>;
```
SebastianMestre, line 13: We should look into putting this in a utils.hpp file in the future

src/test/test_status.hpp
```diff
@@ -0,0 +1,20 @@
...
+	Ok = 0,
+	Error,
+	Fail,
+	Empty,
+};
+
+struct TestReport {
+	test_status m_code;
+	std::string m_msg;
+};
```
SebastianMestre, line 20: That's pretty cool

PR #111, opened 2020-09-01
Add gc handles

SebastianMestre:
This will prevent us from having dangling pointers when we manipulate gc objects

The main machinery is in `interpreter/gc_ptr.hpp`. It uses a tiny bit of template trickery to let us move from a `gc_ptr<subclass>` into a `gc_ptr<superclass>`. I don't know that much about template magic, so I just did the same thing `std::shared_pointer` does in the standard library implementation that's on my machine.

Whether it works or not is TBD, maybe we should try calling `Environment::run_gc` at the top of every `Environment::new_*` function. TBH, my brain is pretty much dry, so we will have to see tomorrow or whenever.

### DIFF
--- a/src/interpreter/environment.cpp
+++ b/src/interpreter/environment.cpp
@@ -114,28 +114,28 @@ gc_ptr<String> Environment::new_string(std::string s)
 { return m_gc->new_string(std::move(s)); }
 
 gc_ptr<Array> Environment::new_list(ArrayType elements)
-{ return m_gc->new_list_unsafe(std::move(elements)); }
+{ return m_gc->new_list(std::move(elements)); }
 
 gc_ptr<Object> Environment::new_object(ObjectType declarations)
-{ return m_gc->new_object_unsafe(std::move(declarations)); }
+{ return m_gc->new_object(std::move(declarations)); }
 
 gc_ptr<Dictionary> Environment::new_dictionary(ObjectType declarations)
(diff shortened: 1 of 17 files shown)


src/interpreter/environment.cpp
```diff
@@ -109,25 +113,25 @@ gc_ptr<Boolean> Environment::new_boolean(bool b)
 gc_ptr<String> Environment::new_string(std::string s)
 { return m_gc->new_string(std::move(s)); }
 
-Array* Environment::new_list(ArrayType elements)
+gc_ptr<Array> Environment::new_list(ArrayType elements)
 { return m_gc->new_list_unsafe(std::move(elements)); }
```
lushoBarlett: Shouldn't you have deleted the "unsafe" methods by this point, and change them for the regular ones?

SebastianMestre: oops they shouldn't be used anymore, but i thought it might be useful to keep them around. I just removed them anyways

SebastianMestre:
> This was... quite the piece of work. Changes propagate everywhere now, it's hard to read it all sometimes. I think there should be one extra commit where you do a final clean up of the "unsafe" functions, but overall its good to go.

Yeah.. This kind of thing happens when you change the return type of an important function. Even though the interface of the type is pretty much the same, updating the call sites can still be necessary

PR #112, opened 2020-09-01
While statement

lushoBarlett:
Implements while loops, in accordance with the task graph.

As a side note, revisiting all those files made me realize that there's a heavy lack in formatting everywhere.

### DIFF
--- a/src/typed_ast.cpp
+++ b/src/typed_ast.cpp
@@ -172,12 +172,12 @@ TypedAST* convertAST(AST::ForStatement* ast) {
 }
 
 TypedAST* convertAST(AST::WhileStatement* ast) {
-    auto typed_for = new WhileStatement;
+    auto typed_while = new WhileStatement;
 
-    typed_for->m_condition = get_unique(ast->m_condition);
-    typed_for->m_body      = get_unique(ast->m_body);
+    typed_while->m_condition = get_unique(ast->m_condition);
+    typed_while->m_body      = get_unique(ast->m_body);
 
-    return typed_for;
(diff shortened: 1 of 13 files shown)


SebastianMestre:
> As a side note, revisiting all those files made me realize that there's a heavy lack in formatting everywhere.

Yeah, I noticed that when touching `native.cpp` on my last pr. It might be cool to run clang-format project-wide and submit it as a pr

src/typed_ast.cpp
```diff
@@ -171,6 +171,15 @@ TypedAST* convertAST(AST::ForStatement* ast) {
     return typed_for;
 }
 
+TypedAST* convertAST(AST::WhileStatement* ast) {
+    auto typed_for = new WhileStatement;
+
+    typed_for->m_condition = get_unique(ast->m_condition);
+    typed_for->m_body      = get_unique(ast->m_body);
+
+    return typed_for;
```
SebastianMestre, line 180: it's a bit ugly that the variable is named `typed_for`. either `typed_while` or `typed_ast` would be more appropriate

lushoBarlett: Woops, missed that one

lushoBarlett:
> > As a side note, revisiting all those files made me realize that there's a heavy lack in formatting everywhere.
> 
> Yeah, I noticed that when touching `native.cpp` on my last pr. It might be cool to run clang-format project-wide and submit it as a pr

I was gonna suggest that exact thing, so if you would like to do it, since you already have it installed, you may proceed.

PR #114, opened 2020-09-02
Typechecker context binding

SebastianMestre:
Hindley-Milner relies on the "context" to know which type variables to generalize, and which ones to leave open for unification.

I implemented this "context' as a part of the CompileTimeEnvironment

### DIFF
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -158,4 +158,27 @@ void CompileTimeEnvironment::exit_function() {
 	m_function_stack.pop_back();
 }
 
+MonoId CompileTimeEnvironment::new_type_var() {
+	MonoId result = m_typechecker.new_var();
+	VarId var = m_typechecker.m_core.mono_data[result].data_id;
+	current_scope().m_type_vars.insert(var);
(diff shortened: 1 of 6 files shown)


SebastianMestre:
Regarding what I described on 3b16bd5a5b93e47df851c9b76515fc831a637641, I have actually verified this is the case.

When you force older vars to point to newer vars, you get over-generalization. In the program below, the type of g is deduced to have 2 quantifiers. However, when you do the opposite, this effect is removed, and it correctly has a single quantifier.

```rust
g := fn(x) {
  f := fn() => x;
  return f();
};
```

SebastianMestre:
> Pretty good catch, I'm a little worried because we are comparing two ints that increase with each monotype, so my DB brain yells ID overflow. But this seems extremely unrealistic... Thoughts?

Remember that monotypes are not persisted, they only exist for the duration of a single run of the frontend. I don't see us processing programs that have `2147483648` (>2000M) monotypes in them any time soon. But if we do hit that, we will have bigger problems, like memory consumption, and execution time.

There are things we could do if that does become a reality, though.

PR #115, opened 2020-09-03
Typechecker declaration graph + rec-rule

SebastianMestre:
I have implemented the rec rule.

~There is no testing so it might not be a good idea to merge it just yet...~

### DIFF
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -181,4 +181,18 @@ bool CompileTimeEnvironment::has_type_var(VarId var) {
 	return scan_scope(m_global_scope, var);
 }
 
+TypedAST::Declaration* CompileTimeEnvironment::current_top_level_declaration() {
+	return m_current_decl;
+}
+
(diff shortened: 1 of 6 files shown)


lushoBarlett:
I don't think I quite get the purpose of this 'hidden type vars' as dummies. Would you mind explaining a little bit to me what that's all about?

SebastianMestre:
> I don't think I quite get the purpose of this 'hidden type vars' as dummies. Would you mind explaining a little bit to me what that's all about?

Hidden type vars are type vars that get generalized no matter what. (Because they are never added to the context)

We give a dummy type var to every declaration in the rec-block because they need to have a type, because they will be referenced by some other declaration within the same rec-block, before they are typechecked.

SebastianMestre:
I just fixed some bugs and added a test. I am pretty confident it works now.

lushoBarlett (approved):
I'm not that close to the type system, but I did _kind of_ understand how this works. Anyhow, the tests actually working are more that enough for me to approve this pr. Great stuff!

PR #116, opened 2020-09-03
Else statement

lushoBarlett:
Implements the else block after an if statement

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -743,9 +743,9 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_return_statement() {
 	return make_writer<std::unique_ptr<AST::AST>>(std::move(e));
 }
 
-Writer<std::unique_ptr<AST::AST>> Parser::parse_if_statement() {
+Writer<std::unique_ptr<AST::AST>> Parser::parse_if_else_statement() {
 	Writer<std::unique_ptr<AST::AST>> result
-	    = { { "Parse Error: Failed to parse if statement" } };
+	    = { { "Parse Error: Failed to parse if-else statement" } };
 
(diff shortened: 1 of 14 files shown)


src/parser.cpp
```diff
@@ -769,11 +782,20 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_if_statement() {
...
 
+	if (consume(token_type::KEYWORD_ELSE)) {
+		auto else_body = parse_statement();
+
+		if (handle_error(result, else_body))
+			return result;
+
+		e->m_else_body = std::move(else_body.m_result);
+	}
+
```
SebastianMestre, line 798: It would be nicer if the parsing of the else body was hoisted to before the creation of the AST node

SebastianMestre (approved):
Looks pretty good. I haven't grepped through the rest of the code to see if you missed anything, but I will assume that's what you did.

src/parser.cpp
lushoBarlett: I think doing this requires that I write the whole type when declaring `else_body`, because I have to call `parse_statement` after checking for `KEYWORD_ELSE` so `auto` wouldn't work. Isn't that worse? Here is the changed code:
```c++
        Writer<std::unique_ptr<AST::AST>> else_body; 

        if (consume(token_type::KEYWORD_ELSE)) {
                else_body = parse_statement();

                if (handle_error(result, else_body))
                        return result;
        }   

        auto e = std::make_unique<AST::IfElseStatement>();

        e->m_condition = std::move(condition.m_result);
        e->m_body = std::move(body.m_result);

        if (else_body.m_result)
                e->m_else_body = std::move(else_body.m_result);
```

SebastianMestre: do as you like, the pr is approved nonetheless

PR #117, opened 2020-09-03
reformat everything

SebastianMestre:
I just ran clang-format. All of the hand-picked formatting is gone, but a lot of the messiness is also gone. meh

I don't love it, tbh.


src/interpreter/gc_ptr.hpp
```diff
@@ -9,27 +9,30 @@ struct gc_ptr {
...
+	template <typename T>
+	friend class gc_ptr;
 
-	template <typename T, typename = typename std::enable_if<std::is_convertible<T*, ValueType*>::value>::type>
+	template <
+	    typename T,
+	    typename = typename std::enable_if<std::is_convertible<T*, ValueType*>::value>::type>
 	gc_ptr(gc_ptr<T>&& o) : m_ptr { o.m_ptr } {
 		o.m_ptr = nullptr;
 	}
```
SebastianMestre, line 32: This is really nice

src/lexer.cpp
```diff
@@ -28,8 +28,12 @@ void Lexer::push_token(token_type tt, int width) {
...
 		text[i] = m_source[m_source_index + i];
 
-	Token t = { tt, std::move(text), m_current_line, m_current_column, m_current_line,
-		m_current_column + width };
+	Token t = { tt,
+		        std::move(text),
+		        m_current_line,
+		        m_current_column,
+		        m_current_line,
+		        m_current_column + width };
```
SebastianMestre, line 36: This I completely despise

lushoBarlett (commented):
Oh I hate it. I don't think the benefits are worth the mess it made on our hand-made formatting. Specially on our header files.

Perhaps our best option is to just fix it as we go along... most issues are on the `.cpp` files and its a matter of spaces to tabs (which can be solved by `vim` instantly) and statements like `if` and `for`

SebastianMestre:
> Oh I hate it. I don't think the benefits are worth the mess it made on our hand-made formatting. Specially on our header files.
> 
> Perhaps our best option is to just fix it as we go along... most issues are on the `.cpp` files and its a matter of spaces to tabs (which can be solved by `vim` instantly) and statements like `if` and `for`

There aren't that many places with custom formatting, tbh. And we could just not run headers through clang-format (or add `/* clang-format off */` where appropiate)

src/lexer.cpp
SebastianMestre: I am pretty sure this is a bug in `clang-format`

SebastianMestre: `Cpp11BracedListStyle: true` should make braced lists format the same than parenthesized lists, but this is not the case.

No matter how much I fiddle with configuration, I still get this style:

```c++

auto t = Token(
	tt,
	std::move(text),
	m_current_line,
	m_current_column,
	m_current_line,
	m_current_column + width);

auto t = Token { tt,
                 std::move(text),
                 m_current_line,
                 m_current_column,
                 m_current_line,
                 m_current_column + width };
```

I want the second declaration to be formatted the same than the first one. but it keeps getting aligned to the right instead of to the left.

SebastianMestre: Yeah. This was caused by a bug in clang-format-9. I upgraded my installation to clang-format-10, and it's good now.

lushoBarlett (approved):
Alright! Get merged son

PR #118, opened 2020-09-04
Typechecker fleshing out

SebastianMestre:
I noticed some of the tests were commented out, the only reason being that we didn't know how to typecheck arrays, and that the built-in array handling functions didn't have types on them. So I went and wrote up what was needed to make the tests run.


lushoBarlett (approved):
No particular comment on this one, we talked about this yesterday. Approved.

PR #119, opened 2020-09-04
Ternary operator

lushoBarlett:
Implements the ternary operator.

~Currently it breaks in the typechecker...~ The Address Sanitizer complains about an illegal access in `typecheck(CallExpression)`, but I can't seem to understand why. I checked if it was a parsing mistake where it thought it was a CallExpression instead of something else, but I figured this should not be the case. I need help to understand if I misused the type system or something.


src/typecheck.cpp
```diff
@@ -200,6 +200,18 @@ void typecheck(TypedAST::IndexExpression* ast, Frontend::CompileTimeEnvironment&
...
+	typecheck(ast->m_condition.get(), env);
+	env.m_typechecker.m_core.unify(
+	    ast->m_condition->m_value_type, env.m_typechecker.mono_boolean());
+
+	typecheck(ast->m_then_expr.get(), env);
+	typecheck(ast->m_else_expr.get(), env);
+
+	env.m_typechecker.m_core.unify(
+	    ast->m_then_expr->m_value_type, ast->m_else_expr->m_value_type);
+}
```
SebastianMestre, line 213: `ast->m_value_type` is never assigned

SebastianMestre (changes requested):

lushoBarlett:
Okay so, I was a little confused as to _why_ something like `i == 0` even worked... After seeing how the `TypedAST` completely ignored it. Then I remembered that every bin-op is translated into a native function, so then it hit me why the failure ocurrs in a `CallExpression`. It's the `+` *function* that's getting called.

I probably need to check out where in the codebase that transformation happens, because I don't remember.

SebastianMestre:
The code looks good to me, and the tests pass... I'm not quite clear on whether it works, though. Your PR description seems to imply it doesn't. Does it work?

SebastianMestre:
Ok that's all cleared up, then

SebastianMestre (approved):
I have no comments, good job

lushoBarlett:
Just to leave record, the code **works**. I just left the problem (and the discovery of the problem) as a reminder of what happened. Also I felt like dashing all of that text would look bad.

PR #120, opened 2020-09-05
TypeChecker polish

SebastianMestre:
Some small touch ups that make the TC a bit nicer.

### DIFF
--- a/src/test/main.cpp
+++ b/src/test/main.cpp
@@ -426,6 +426,21 @@ void interpreter_tests(Test::Tester& tests) {
 	    +[](Interpreter::Environment& env) -> exit_status_type {
 		    return Assert::equals(eval_expression("__invoke()", env), 2);
 	    }));
+
+	// we really only care that it typechecks, but might as
+	// well throw a whole batch of stuff at it.
+	tests.add_test(std::make_unique<Test::InterpreterTestSet>(
+	    R"(
+			__invoke := fn() {
+				fib:= fn(n) => (if n < 2
+					then n
+					else fib(n-1) + fib(n-2));
+				return fib(3);
+			};
+		)",
+	    Testers {+[](Interpreter::Environment& env) -> exit_status_type {
+		    return Assert::equals(eval_expression("__invoke()", env), 2);
+	    }}));
 }
 
 void tarjan_algorithm_tests(Test::Tester& tester) {
--- a/src/compile_time_environment.cpp
+++ b/src/compile_time_environment.cpp
@@ -83,6 +83,15 @@ CompileTimeEnvironment::CompileTimeEnvironment() {
 			auto poly_id = m_typechecker.m_core.new_poly(term_mono_id, {});
 			declare_builtin("array_join", poly_id);
 		}
+		{
+			auto term_mono_id = m_typechecker.m_core.new_term(
+			    TypeChecker::BuiltinType::Function,
+			    {array_mono_id, m_typechecker.mono_int(), var_mono_id},
+			    "[builtin] (array(<a>), int) -> a");
+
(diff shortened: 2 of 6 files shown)


SebastianMestre:
very happy, this typechecks and runs to completion, now.
```rust
map := fn(arr, f){
	res := array{};
	for(i := 0; i < size(arr); i = i+1)
		array_append(res, f(array_at(arr, i)));
	return res;
};

filter := fn(arr, f){
	res := array{};
	for(i := 0; i < size(arr); i = i+1)
		if(f(array_at(arr, i)))
			array_append(res, array_at(arr, i));
	return res;
};

__invoke := fn(){
	return array{1;2;3;4;5;6;7;}
		|> map(fn(x) => x*3)
		|> filter(fn(x) => x < 10);
};
```

src/test/main.cpp
```diff
@@ -426,6 +426,21 @@ void interpreter_tests(Test::Tester& tests) {
...
 	    }));
+
+	// we really only care that it typechecks, but might as
+	// well throw a whole batch of stuff at it.
+	tests.add_test(std::make_unique<Test::InterpreterTestSet>(
+	    R"(
+			__invoke := fn() {
+				fib:= fn(n) => (if n < 2
+					then n
+					else fib(n-1) + fib(n-2));
```
lushoBarlett, line 437: I'm very happy that we chose `if ... then ... else ...` as the ternary operator syntax. It looks amazing

lushoBarlett (approved):
Its nice to be able to work with arrays, although I'm very much against the `array_at` function to be definitive, since this is a job for `[...]`. Furthermore, all native array functions in general have long and stupid names in my opinion.

We will need a way to help the programmer not write that much just to append or join two arrays. Perhaps this can be achieved in the future with syntactic sugar and shorter (and better) function names.

SebastianMestre:
> Its nice to be able to work with arrays, although I'm very much against the `array_at` function to be definitive, since this is a job for `[...]`.

well, an index operator that acts as both `(array(<a>), Int)->a` and `(dictionary(<k,a>), k)->a` is a bit far away in terms of the type system. You could look into it, though.

> Furthermore, all native array functions in general have long and stupid names in my opinion.

You can just re-bind any of those functions at any time, just do `push := array_append;`.

> We will need a way to help the programmer not write that much just to append or join two arrays.

I don't see this as an issue. `array_expand(arr1, arr2)` is a fine way to concat two arrays

> Perhaps this can be achieved in the future with syntactic sugar and shorter (and better) function names.

I don't think it matters too much what the functions are named, as long as it is somewhat reasonable (as I think it is now).

I have thought about these issues, and I think the solution is not coming up with shorter names, but rather implement a more general (maybe hierarchical) mechanism for organizing names.

Consider the following piece of code.

```rust
__invoke := fn () {
    arr1 := array{};
    arr2 := array{};
    array_append(arr1, 3);
    array_append(arr2, 3);
    return array_extend(arr1, arr2);
};
```
Now, look at a version using a possible module feature, where `import` returns a namespace with all the declarations in a file.

```rust
__invoke := fn () {
    Array := import("array_utils.jp");
    arr1 := array{};
    arr2 := array{};
    Array.append(arr1, 3);
    Array.append(arr2, 3);
    return Array.extend(arr1, arr2);
};
```

Then, this could be made nicer like this, where `embed` pulls all the declarations of the given namespace into the current scope.

```rust
__invoke := fn () {
    embed import("array_utils.jp");
    arr1 := array{};
    arr2 := array{};
    append(arr1, 3);
    append(arr2, 3);
    return extend(arr1, arr2);
};
```

"embed" is essentially a form of declaration, so it could appear at either file scope, block scope, or, possibly, module scope.

PR #121, opened 2020-09-05
Docs improvement

SebastianMestre:
I went and touched up some of the docs. There were some spelling mistakes and such, but also, they were starting to get outdated.

I want to update the code example on the readme, but I don't know what would be a nice thing to replace the one we currently have. any ideas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,20 +2,39 @@
 
 We are trying to make a nice programming language which is
 
- - easy to refactor
- - consistent
- - nice to use
+ - Nice to use
+ - Easy to refactor
+ - Consistent
+
+For this purpose, Jasper has:
+
+ - Type deduction
+ - First class functions
+ - Consistent and context-free syntax
+ - Syntactic sugar
+ - Many others
 
 Here is an example piece of code:
 
 ```c++
-f := fn (name) {
-	greeting := "Hello, ";
-	return greeting + name;
+or_default := fn(str, default) =>
+	(if str == ""
+		then default
+		else str);
+
+make_greeting := fn(prefix, suffix) =>
+	fn(name) =>
+		prefix + name + suffix;
+
+say_hello := fn(name) {
+	greeting := make_greeting("Hello, ", "!");
+	return name
+		|> or_default("Friend")
+		|> greeting();
 };
 
-__invoke := fn () {
-	print(f("Programmer"));
+__invoke := fn() {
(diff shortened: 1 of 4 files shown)


lushoBarlett:
I mean, if you intend to use unfinished features, definitely use objects. You should also use arrow functions. And the example could be just an object with some data (typed variables of course) and just pass it to some function to gather and print the data it contains. For this you may even want to use the `|>` operator.

As such you illustrate all the mayor features of the language, plus the notion that Data Oriented Programming is possible and effective.

Perhaps this is too difficult and confusing for a new user, in which case you could just portray a short algorithm (preferably not abusing array functions because we did not polish their name and use).

SebastianMestre:
> I mean, if you intend to use unfinished features

I really want the examples to be able to be run by whoever wishes to download the interpreter

> ... use arrow functions. ... pass it to some function to gather and print the data it contains. ... use the `|>` operator.

Yeah, I was more asking for a piece of code that does these things

> plus the notion that Data Oriented Programming is possible and effective.

Not really sure what you mean by "Data Oriented Programming", the thing I know as DOD is related to controlling the memory layout of your data, which is 100% not a thing in Jasper.

lushoBarlett:
I think of DOP as basically having data on one hand, and functions that transform the data on the other. Instead of data and functions clumped up in a class which is what happens with OOP.

As for the example:
```rust
decorator :=
    fn(prefix, suffix) =>
        fn(x) =>
            prefix + x + suffix;

__invoke() := fn() {
    salute := decorator("Hello ", "!");
    print(salute("World"));
};
```

lushoBarlett:
Second version with `|>` operator if that's what you want:
```rust
decorator :=
    fn(prefix, suffix) =>
        fn(x) =>
            prefix + x + suffix;

default_entity :=
    fn(entity) =>
        (if entity == "" then "World" else entity);

__invoke := fn() {
    entity := "";
    salute := decorator("Hello ", "!");
    return entity
               |> default_entity()
               |> salute();
};
```

SebastianMestre:
Hey, those are really nice examples! I have changed them up to my liking and put them in the README.

SebastianMestre:
I'm happy with how this looks, so I would appreciate it if you reviewed the pr.

lushoBarlett (approved):
It's very cool

PR #122, opened 2020-09-07
General cleanup

SebastianMestre:
I started doing some general cleanup and refactoring, and ended up making match_identifiers return an error report object. talk about scope creep, huh.

`typecheck` should also return an ErrorReport. So I do it in this pr, or do I leave it for later?

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -103,6 +103,14 @@ struct FunctionLiteral : public AST {
 	    : AST {ast_type::FunctionLiteral} {}
 };
 
+struct ShortFunctionLiteral : public AST {
+	std::unique_ptr<AST> m_body;
+	std::vector<std::unique_ptr<AST>> m_args;
+
+	ShortFunctionLiteral()
+	    : AST {ast_type::ShortFunctionLiteral} {}
+};
+
 struct DeclarationList : public AST {
 	std::vector<std::unique_ptr<AST>> m_declarations;
 
--- a/src/match_identifiers.cpp
+++ b/src/match_identifiers.cpp
@@ -4,25 +4,37 @@
 #include <unordered_map>
 
 #include "compile_time_environment.hpp"
+#include "error_report.hpp"
 #include "typed_ast.hpp"
 
 #include <cassert>
 
 namespace TypeChecker {
 
-void match_identifiers(
+#define CHECK_AND_RETURN(expr)                                                 \
+	{                                                                          \
+		auto err = expr;                                                       \
+		if (!err.ok())                                                         \
+			return err;                                                        \
+	}
+
+[[nodiscard]] ErrorReport match_identifiers(
     TypedAST::Declaration* ast, Frontend::CompileTimeEnvironment& env) {
 
 	ast->m_surrounding_function = env.current_function();
 	env.declare(ast->identifier_text(), ast);
 
 	if (ast->m_value)
-		match_identifiers(ast->m_value.get(), env);
+		CHECK_AND_RETURN(match_identifiers(ast->m_value.get(), env));
+
+	return {};
 }
 
-void match_identifiers(TypedAST::Identifier* ast, Frontend::CompileTimeEnvironment& env) {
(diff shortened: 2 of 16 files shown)


lushoBarlett:
I mean... You already did it for match so... yeah do it. It's not a particularly complicated thing to do. It's just wrapping a return value into a type with a string in it, which we would have to do anyway at some point.

src/ast.hpp
```diff
@@ -103,6 +103,14 @@ struct FunctionLiteral : public AST {
...
 };
 
+struct ShortFunctionLiteral : public AST {
+	std::unique_ptr<AST> m_body;
+	std::vector<std::unique_ptr<AST>> m_args;
+
+	ShortFunctionLiteral()
+	    : AST {ast_type::ShortFunctionLiteral} {}
+};
+
```
lushoBarlett, line 113: What is the benefit of this?

src/match_identifiers.cpp
```diff
@@ -4,25 +4,37 @@
...
 
-void match_identifiers(
+#define CHECK_AND_RETURN(expr)                                                 \
+	{                                                                          \
+		auto err = expr;                                                       \
+		if (!err.ok())                                                         \
+			return err;                                                        \
+	}
+
+[[nodiscard]] ErrorReport match_identifiers(
```
lushoBarlett, line 21: `[[nodiscard]]` is a C++17 attribute. Although very useful, using it would mean that we should be discarding our home made `string_view.xpp`, because its main purpose was to lower us to C++14. We should decide on that.

src/match_identifiers.cpp
```diff
@@ -96,50 +112,69 @@ void match_identifiers(
 	// scan body
 	assert(ast->m_body->type() == ast_type::Block);
 	auto body = static_cast<TypedAST::Block*>(ast->m_body.get());
-	for (auto& child : body->m_body)
-		match_identifiers(child.get(), env);
+	for (auto& child : body->m_body) {
+		CHECK_AND_RETURN(match_identifiers(child.get(), env));
+	}
```
lushoBarlett, line 118: Nit: the for loop braces here are not consistent with other for loops in the diff

lushoBarlett (commented):
One suggestion for cleanup that I would like for you to check out, if possible, is the `Writer` struct in `parser.hpp` (I think).

src/ast.hpp
SebastianMestre: Less ad-hoc code in the parser, it would be helpful in writing tools, like a code formatter.

src/match_identifiers.cpp
SebastianMestre: Not really. attributes are ignored by compilers that don't understand them. And some may provide the feature, even when compiling with older standards.

SebastianMestre:
> I mean... You already did it for match so... yeah do it. It's not a particularly complicated thing to do. It's just wrapping a return value into a type with a string in it, which we would have to do anyway at some point.

Well, one would actually have to splice this thing into TypeSystemCore, and TypeChecker, because that's where errors come from.

But since some of those functions actually return things, instead, one should return something like a `Writer`

src/match_identifiers.cpp
lushoBarlett: I did not know that, huh

src/ast.hpp
lushoBarlett: Fair enough

lushoBarlett:
> But since some of those functions actually return things, instead, one should return something like a Writer, which is a bit more involved

If you want to do it, I'm not against it being in this pr, but it's up to you.

SebastianMestre:
> > But since some of those functions actually return things, instead, one should return something like a Writer, which is a bit more involved
> 
> If you want to do it, I'm not against it being in this pr, but it's up to you.

I'm gonna leave it here for now, this pr is big enough

lushoBarlett (approved):
Overall nice and needed changes

PR #124, opened 2020-09-08
TSC enhancement

lushoBarlett:
To bind types to names, I'm thinking of something of the sorts of `x @= <type>`. Why? Mainly because there was a time where we were discussing type functions and the `@type{ ... }` notation came up. If we would continue on the path of types in the Jasper source code, we **could** consider using the `@` character for our needs.

I believe this might be a competent idea over others because this syntax can be distinguished from the rest of the declarations, and overall... things. _Although_ I don't know if this is actually the best way to differentiate it from everything else (or if we want to do that), but it's a way. This pr will go slowly as we choose how to handle this.

Example use:
```rust
predicate @= x -> Boolean;
iter @= array{} | myIterType{};

filter_iterable : iter predicate -> iter = fn (array, pred) { ... };
```
Ignore the fact that I assumed a bunch of syntax, although a very interesting thing to think about.

### DIFF
--- a/src/token_type.hpp
+++ b/src/token_type.hpp
@@ -26,7 +26,6 @@ constexpr const char* token_type_string[] = {
 
 	"=",
 	":=",
-	"@=",
 	":",
 
 	"INCREMENT",
@@ -107,7 +106,6 @@ enum class token_type {
 
 	ASSIGN,
 	DECLARE_ASSIGN,
-	DECLARE_TYPE,
 	DECLARE,
 
 	INCREMENT,
--- a/src/token_type.hpp
+++ b/src/token_type.hpp
@@ -26,6 +26,7 @@ constexpr const char* token_type_string[] = {
 
 	"=",
 	":=",
+	"@=",
 	":",
 
 	"INCREMENT",
@@ -106,6 +107,7 @@ enum class token_type {
 
 	ASSIGN,
 	DECLARE_ASSIGN,
+	DECLARE_TYPE,
 	DECLARE,
 
 	INCREMENT,
--- a/src/typechecker.cpp
+++ b/src/typechecker.cpp
@@ -7,14 +7,14 @@
 namespace TypeChecker {
 
 TypeChecker::TypeChecker() {
-	m_core.type_function_data.push_back({-1}); // 0 | function
-	m_core.type_function_data.push_back({0});  // 1 | int
-	m_core.type_function_data.push_back({0});  // 2 | float
-	m_core.type_function_data.push_back({0});  // 3 | string
-	m_core.type_function_data.push_back({1});  // 4 | array
-	m_core.type_function_data.push_back({1});  // 5 | dictionary
-	m_core.type_function_data.push_back({0});  // 6 | boolean
-	m_core.type_function_data.push_back({0});  // 7 | unit
+	m_core.type_function_data.push_back({-1, type_function_type::Known}); // 0 | function
+	m_core.type_function_data.push_back({0, type_function_type::Known});  // 1 | int
+	m_core.type_function_data.push_back({0, type_function_type::Known});  // 2 | float
+	m_core.type_function_data.push_back({0, type_function_type::Known});  // 3 | string
+	m_core.type_function_data.push_back({1, type_function_type::Known});  // 4 | array
+	m_core.type_function_data.push_back({1, type_function_type::Known});  // 5 | dictionary
+	m_core.type_function_data.push_back({0, type_function_type::Known});  // 6 | boolean
+	m_core.type_function_data.push_back({0, type_function_type::Known});  // 7 | unit
 
 	m_core.term_data.push_back({1, {}, "builtin int"}); // 0 | int(<>)
 	m_core.term_data.push_back({2, {}, "builtin float"}); // 1 | float(<>)
--- a/src/typechecker.cpp
+++ b/src/typechecker.cpp
@@ -7,40 +7,38 @@
 namespace TypeChecker {
 
 TypeChecker::TypeChecker() {
-	m_core.type_function_data.push_back({-1, type_function_type::Known}); // 0 | function
-	m_core.type_function_data.push_back({0, type_function_type::Known});  // 1 | int
-	m_core.type_function_data.push_back({0, type_function_type::Known});  // 2 | float
-	m_core.type_function_data.push_back({0, type_function_type::Known});  // 3 | string
-	m_core.type_function_data.push_back({1, type_function_type::Known});  // 4 | array
-	m_core.type_function_data.push_back({1, type_function_type::Known});  // 5 | dictionary
-	m_core.type_function_data.push_back({0, type_function_type::Known});  // 6 | boolean
-	m_core.type_function_data.push_back({0, type_function_type::Known});  // 7 | unit
+	m_core.new_type_function(-1); // 0 | function
+	m_core.new_type_function(0);  // 1 | int
+	m_core.new_type_function(0);  // 2 | float
+	m_core.new_type_function(0);  // 3 | string
+	m_core.new_type_function(1);  // 4 | array
+	m_core.new_type_function(1);  // 5 | dictionary
+	m_core.new_type_function(0);  // 6 | boolean
+	m_core.new_type_function(0);  // 7 | unit
 
-	m_core.term_data.push_back({1, {}, "builtin int"}); // 0 | int(<>)
-	m_core.term_data.push_back({2, {}, "builtin float"}); // 1 | float(<>)
-	m_core.term_data.push_back({3, {}, "builtin string"}); // 2 | string(<>)
-	m_core.term_data.push_back({6, {}, "builtin bool"}); // 3 | boolean(<>)
-	m_core.term_data.push_back({7, {}, "builtin unit"}); // 4 | unit(<>)
-
-	m_core.mono_data.push_back({mono_type::Term, 0}); // 0 | int(<>)
-	m_core.mono_data.push_back({mono_type::Term, 1}); // 1 | float(<>)
-	m_core.mono_data.push_back({mono_type::Term, 2}); // 2 | string(<>)
-	m_core.mono_data.push_back({mono_type::Term, 3}); // 3 | boolean(<>)
-	m_core.mono_data.push_back({mono_type::Term, 4}); // 4 | unit(<>)
+	m_core.new_term(1, {}, "builtin int"); // 0 | int(<>)
+	m_core.new_term(2, {}, "builtin float"); // 1 | float(<>)
+	m_core.new_term(3, {}, "builtin string"); // 2 | string(<>)
+	m_core.new_term(6, {}, "builtin bool"); // 3 | boolean(<>)
+	m_core.new_term(7, {}, "builtin unit"); // 4 | unit(<>)
 }
 
 MonoId TypeChecker::mono_int() {
 	return 0;
 }
+
 MonoId TypeChecker::mono_float() {
 	return 1;
 }
+
 MonoId TypeChecker::mono_string() {
 	return 2;
 }
+
 MonoId TypeChecker::mono_boolean() {
 	return 3;
 }
+
 MonoId TypeChecker::mono_unit() {
 	return 4;
 }
--- a/src/typesystem.cpp
+++ b/src/typesystem.cpp
@@ -274,3 +274,41 @@ void TypeSystemCore::func_unify(TypeFunctionId a, TypeFunctionId b) {
 	else
 		assert(0 and "unifying two differente known type functions");
 }
+
+TypeVarData TypeSystemCore::var_find(TypeVarId type_var) {
+	TypeVarData& var_data = type_vars[type_var];
+
+	switch(var_data.kind) {
+	case kind_type::Mono:
+		return {kind_type::Mono, find(var_data.type_var_id)};
+	case kind_type::Poly:
+		return {kind_type::Poly, var_data.type_var_id};
+	case kind_type::TypeFunction:
+		return {kind_type::TypeFunction, func_find(var_data.type_var_id)};
+	}
+
+	assert(0 and "unknown kind");
+}
+
+void TypeSystemCore::var_unify(TypeVarId a, TypeVarId b) {
+	TypeVarData rep_a = var_find(a);
+	TypeVarData rep_b = var_find(b);
+
+	assert(rep_a.kind == rep_b.kind and "cannot unify different kinds");
+
+	if (rep_a.type_var_id == rep_b.type_var_id)
+		return;
+
+	switch(rep_a.kind) {
+	case kind_type::Mono:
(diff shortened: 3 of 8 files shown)


SebastianMestre:
What i had in mind was to just use `:=` for declaring types, and @ to indicate type vars

Type terms currently use `type_func(<arg1, arg2>)`

```
predicate := @x -> boolean;
iter := array(<@x>) | myIter;
filter : (iter, predicate) => iter = fn(seq, pred) { ... };
```

lushoBarlett:
> What i had in mind was to just use := for declaring types

I thought that maybe using `:=` would mean that types can themselves have types (kinds) and well, now that I think about it, that does sound like something we might want.

> Multiple types is probably easy to do. (mutually recursive types, e.g. an AST)

I'm not sure what you mean by this, are you referring to type inference? If so, yea, it's not a problem.

> A mix of types and values is weird, maybe it should be illegal, at least for now.

I'm sure I don't know what you mean by this.

SebastianMestre:
> > Multiple types is probably easy to do. (mutually recursive types, e.g. an AST)
> 
> I'm not sure what you mean by this, are you referring to type inference? If so, yea, it's not a problem.

Look at this Haskell-style datatype declaration

```hs
data TreeList = Nil | Cons Tree TreeList
data Tree = Node Int TreeList
```
Since the types are mutually recursive, when we process one of the declarations, we would see an identifier that refers to a type function (either `Tree` or `TreeList`) that has not yet been processed. How would this be handled in the typechecker?

It looks to me like we need some way to leave types partially incomplete, and then finish them off.

One way to do this is with unification (just like we do with monotypes of values). We put a dummy typefuncvar on every declaration, and then unify them as we go along.

Another possibility is to do an extra pass. First build up what the type is, marking some (possibly all) of its data as not-known-yet, then do a second pass where we fill those out.

> > A mix of types and values is weird, maybe it should be illegal, at least for now.
> 
> I'm sure I don't know what you mean by this.

Well, what would happen if we have a value that is referenced by a type? Is this legal? Do we error out as soon as we detect it?

```
pepe := popo { ... something ... }; // pepe is of type popo
popo := ... something with `pepe` ...; // popo is type, but pepe is referenced somewhere in its definition
```

lushoBarlett:
> Also, referencing non-types in types is ~probably~ bad.

You mean definitely.

Unless you can show me an example of it being used in some other language, in conjunction with us liking it.

Also, I believe we could abuse the typesystem to work for type vars as well, so we don't have to "do another pass". I think we should reuse that.

SebastianMestre:
> > Also, referencing non-types in types is ~probably~ bad.
> 
> You mean definitely.
> 
> Unless you can show me an example of it being used in some other language, in conjunction with us liking it.

Dependent types are all the rage nowadays. You can check out Idris and Idris 2 for some cool examples. (dependent types are kinda hard to implement, though, so we probably want to hold off on that)

https://en.wikipedia.org/wiki/Category:Dependently_typed_languages

> Also, I believe we could abuse the typesystem to work for type vars as well, so we don't have to "do another pass". I think we should reuse that.

Maybe, I don't know how much of the tc core would need to be changed (I haven't given it much thought yet), but being able to extend it, instead of a full re-write, would be good.

I think in this case "another pass" just means a second for loop after the inner loop of `typecheck(DeclarationList)`

lushoBarlett:
The dependent type stuff seems... weird. I don't really know what to say other than _"I don't understand it"_. But it seems really difficult to do.

SebastianMestre:
Would you like to implement type func vars in the tc core?

lushoBarlett:
> Would you like to implement type func vars in the tc core?

I will try. Should I close this pr and start another one fresh to do that?

SebastianMestre:
> > Would you like to implement type func vars in the tc core?
> 
> I will try. Should I close this pr and start another one fresh to do that?

i think it's ok either way

lushoBarlett:
I don't know what else to add here. I feel like any other function regarding type variables should be somewhere else.

src/typesystem.hpp
```diff
@@ -55,6 +55,15 @@ struct PolyData {
...
 };
 
+// RENAME: 'Kind' isn't the exactly the right terminology here
+enum class kind_type { TypeFunction, Mono, Poly };
+
+// Variable that can contain any type, and even type functions
+struct TypeVarData {
+	kind_type kind;
+	TypeVarId type_var_id;
+};
```
SebastianMestre, line 65: Huh, that's interesting. What's it for?

SebastianMestre: Oh, actually, I think I see what this is for. good foresight!

SebastianMestre (commented):
you still didn't add type func vars, though (EDIT: let me be clear with what I mean: TypeFunctionData should be more like MonoData, in that it allows an actual concrete value (a Term in the case of MonoData), or a variable. This will be useful when we start building recursive data types)

also, you should add a way to `unify()` these

src/typesystem.hpp
lushoBarlett: It's for `x := <type>`, I don't know if it can be used in some other way

SebastianMestre: yeah, it will probably be useful to be able to handle any type-like thing with a uniform interface, without having to bloat CompileTimeEnvironment with different methods to handle monotypes, polytypes, and type funcs

SebastianMestre:
Actually, would you like to go all the way and implement either structs, tuples, or unions? I won't have much time in the next few days, (i got ioi2020 coming up) and it would be cool if you did it in the meantime.

lushoBarlett:
> you still didn't add type func vars, though

Then I don't think I'm understanding what I have to do. Do you mean stuff like `@x`? I wanted to ask you about that, because I don't know the full spec of it. What I think it should be/do is just a free var, bound to a type variable (or inlined in a declaration type signature?).

So for example:
```rust
// is this x bound to global or just to iter?
// is it then defined when first seen?
iter := array(<@x>);

// is this x "already defined"
pred := x => bool;

filter : (iter, pred) => iter = fn(arr, p) {
    answer : iter = array{};
    // where is y bound to?
    // it should unify with whatever x means
    elem : @y;
    for (int i := 0; i < arr |> array_size(); i = i + 1) {
        elem = arr |> array_at(i);
        if (elem |> p())
            answer |> array_append(elem);
    }
    return answer;
};
```

lushoBarlett:
> Actually, would you like to go all the way and implement either structs, tuples, or unions?

I can certainly try. I'll give it a shot.

SebastianMestre:
right now we only have monotype vars. I think that making it so that typefuncs can be vars would be good.

That is: TypeFunctionData should be more like MonoData, in that it allows an actual concrete value (a Term in the case of MonoData), or a variable. This will be useful when we start building recursive data types

SebastianMestre:
What you implemented is a way to refer to any type-ish thing(mono type, poly type, or type function).

Instead, there should be vars that are type functions.

lushoBarlett:
> Instead, there should be vars that are type functions.

Ah, so you mean that we introduce a mechanism, such that we can also deduce and unify type functions (because they may be unknown).

SebastianMestre:
In my mind, `@x` should mean "create a monotype var and bind it to the name `@x`" (i.e. call `new_var`, and do some stuff with the CompileTimeEnvironment)

----------------------------

Now, what did I mean with type func vars?

### The problem

The thing with type function vars is that, sometimes, we could construct a type term, but still dont know what the type function is.

If you don't believe me, i'm going to walk through an example of what would happen when we typecheck some mutually-recursive type definitions.

Here is the hypothetical piece of code we will be taking a look at:

```
TreeList := struct {
  values : array(<Tree>);
};

Tree := struct {
  value : Int;
  children : TreeList;
};
```

step 1 : gather_identifiers
result:

TreeList references: array, Tree.
Tree references: int, TreeList.

step 2: generate rec blocks (using tarjan solver)
result:

There is one rec block. its members are: Tree, TreeList

step 3: typecheck the block.
step 3.1: typecheck `TreeList`
step 3.1.1: recurse into `struct` expression.
step 3.1.2: recurse into right hand of `values : array(<Tree>);`
step 3.1.3: type of `values` is:
```c++
Term {
  .type_function = array,
  .arguments = {
    Term {
      .type_function = /* whatever "Tree" is, not yet known, but it must be a type function */,
     .arguments = {}
    }
  }
}
```
step 3.1.4: `TreeList` is
```c++
StructTypeFunction {
  .members = {
    {"values", Term { array, { Term { /* whatever Tree is (UNKNOWN) */, {} } } } }
  }
}
```

step 3.2: typecheck `Tree`
step 3.2.1: recurse into `struct` expression.
step 3.2.2: recurse into right hand of `value : int`
step 3.2.3: type of `value` is:
```c++
Term {
  .type_function = int,
  .arguments = {}
}
```
step 3.2.4: recurse into right hand of `children : TreeList`
step 3.2.5: type of `children` is:
```c++
Term {
  .type_function = /* whatever TreeList is, which we do know, at this point */,
  .arguments = {}
}```
step 3.2.5: `Tree` is:
```c++
StructTypeFunction {
  .members = {
    {"value", Term { int, {} } }
    {"children", Term { /* whatever TreeList is */, {} } }
  }
}
```

At the end, we have


```c++
TreeList = StructTypeFunction {
  .members = {
    {"values", Term { array, { Term { UNKNOWN, {} } } } }
  }
};

Tree = StructTypeFunction {
  .members = {
    {"value", Term { int, {} } }
    {"children", Term { /* whatever TreeList is */, {} } }
  }
};
```

As you can see, this breaks, because we need to build type terms (which, as a reminder, are a form of monotype), of which we don't yet know the type function.

### the solutions:

To fix this, we need to do one of the following:
 - go over all the type declarations once more, filling out all the UNKNOWNs
 - Use unification to fill the UNKNOWNs as we go along
 - something else that I have not though of

with both solutions, we need to do an extra pass.

In the case of the first solution, we need to walk the ASTs again, and fill out the unknowns that way. This is kinda finicky, because we will end up with a bunch of incomplete types in the meantime, so there will probably be a lot of if statements.

In the case of the second one, first, we need to give dummy typefunc vars to every declaration (just like with normal rec blocks, but using typefunc vars instead of monotype vars). Then, as we typecheck each decl, we unify its result with the variable we had assigned to it. After that, the type func has suddenly popped into a bunch of type terms whose typefunc was previously unknown. because it was previously unknown, we couldn't check that we were passing the right number of args, so we need to do that now.

SebastianMestre:
> > Instead, there should be vars that are type functions.
> 
> Ah, so you mean that we introduce a mechanism, such that we can also deduce and unify type functions (because they may be unknown).

exactly

lushoBarlett:
Also, now I understand why you build "dummy" type functions for all the 0 argument types, it's easier on the base case of parsing a monotype, since there is no special case for trivial 0 argument monotypes, nice thinking.

lushoBarlett:
I will add unification tomorrow!

src/typesystem.hpp
```diff
@@ -34,7 +36,7 @@ struct MonoData {
 };
 
 // A variable is just a name for a different monotype.
-// VarData stores a MonoID that indicates which monotype it is equal
+// VarData stores a MonoId that indicates which monotype it is equal
 // to. It can also indicate that it is equal to itself, meaning that
 // we don't know its concrete type.
 struct VarData {
```
SebastianMestre, line 42: You might want to merge master into your branch before you get too many conflicts

lushoBarlett:
At this point this is so ugly that I kinda feel like we should merge here and refactor the type system. Primarily replace the type aliases for something type safe and overload stuff. Opinion?

src/typesystem.cpp
```diff
@@ -246,3 +246,31 @@ MonoId TypeSystemCore::inst_fresh(PolyId poly) {
...
+
+	TypeFunctionData& rep_a = type_function_data[a];
+	TypeFunctionData& rep_b = type_function_data[b];
+
+	if (rep_a.type == type_function_type::Var)
+		rep_a.equals = b;
+	else if (rep_b.type == type_function_type::Var)
+		rep_b.equals = a;
+	else
+		assert(0 and "unifying two differente known type functions");
```
SebastianMestre, line 275: typo (differente)

src/typesystem.cpp
```diff
@@ -246,3 +246,31 @@ MonoId TypeSystemCore::inst_fresh(PolyId poly) {
 		vals.push_back(new_var());
 	return inst_with(poly, vals);
 }
+
+TypeFunctionId TypeSystemCore::func_find(TypeFunctionId func) {
+	TypeFunctionData& func_data = type_function_data[func];
+
+	if (func_data.equals == func or
+	    func_data.type == type_function_type::Known)
```
SebastianMestre, line 254: NIT: this check should be backwards.

I know it is equivalent, but checking `.equals` when the func is known makes no sense.

src/token_type.hpp
```diff
@@ -26,6 +26,7 @@ constexpr const char* token_type_string[] = {
 
 	"=",
 	":=",
+	"@=",
```
SebastianMestre, line 29: this should probably be reverted

src/typechecker.cpp
```diff
@@ -7,14 +7,14 @@
...
 TypeChecker::TypeChecker() {
-	m_core.type_function_data.push_back({-1}); // 0 | function
-	m_core.type_function_data.push_back({0});  // 1 | int
-	m_core.type_function_data.push_back({0});  // 2 | float
-	m_core.type_function_data.push_back({0});  // 3 | string
-	m_core.type_function_data.push_back({1});  // 4 | array
-	m_core.type_function_data.push_back({1});  // 5 | dictionary
-	m_core.type_function_data.push_back({0});  // 6 | boolean
-	m_core.type_function_data.push_back({0});  // 7 | unit
+	m_core.type_function_data.push_back({-1, type_function_type::Known}); // 0 | function
```
SebastianMestre, line 10: NIT: This could be a method of `TypeSystemCore`

SebastianMestre (approved):
> At this point this is so ugly that I kinda feel like we should merge here and refactor the type system. Primarily replace the type aliases for something type safe and overload stuff. Opinion?

I think it's fine as it is, it probably just seems ugly to you because you haven't gotten used to the style, and don't have this part of the codebase mapped out in your head yet.

Also, here, the type functions should get unified.

https://github.com/SebastianMestre/Jasper/blob/f5b8512fb05a06053de39efda031f1aefa7e98c9/src/typesystem.cpp#L185-L191

-------------------

As an aside, (and this will be helpful when known type functions store a lot more data) I think `TypeFunctionData` should look even more like `MonoData`: just a tag and an index, and the data (currently just the argument count) stored in another array. Maybe not in this pr or the next, but it is something to keep in mind for the future

src/typechecker.cpp
lushoBarlett: Everything there could actually

lushoBarlett:
I require a re-review because of commit fd315fa

SebastianMestre (approved):
Looks good to me

PR #125, opened 2020-09-08
Tc core simplification

SebastianMestre:
It turns out, type vars don't have any data to them, so we don't really need to put them in a whole new array, we can just use the `data_id` in `MonoData` as the var index.

This is a lot simpler, and more efficient.

PS: I also added or touched up some comments

### DIFF
--- a/src/typesystem.hpp
+++ b/src/typesystem.hpp
@@ -23,7 +23,7 @@ enum class mono_type { Var, Term };
 // A monotype is a reference to a concrete type. It can be a
 // variable or a term.
 // We express this variant using an enum, and an index that points
-// to the where the data is in the correct TypeSystemCore vector.
+// to where the data is in the correct TypeSystemCore vector.
 //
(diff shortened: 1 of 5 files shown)


src/typesystem.hpp
```diff
@@ -20,14 +20,17 @@ struct TypeFunctionData {
...
-// or a term.
+// A monotype is a reference to a concrete type. It can be a
+// variable or a term.
 // We express this variant using an enum, and an index that points
-// to the where the concrete data is in the correct TypeSystemCore
-// vector.
-// If type is mono_type::Var, then the data is stored in
-// TypeSystemCore::var_data. And if it is mono_type::term, the data
-// is stored in TypeSystemCore::term_data.
+// to the where the data is in the correct TypeSystemCore vector.
```
lushoBarlett, line 26: typo

lushoBarlett (approved):
I don't believe this conflicts with our ability to conveniently store type metadata later, so this is nice.

PR #126, opened 2020-09-12
New builtin types

lushoBarlett:
Short pr, because what I want to do next is related to parsing types, and it will be preparation for type variables actually.

### DIFF
--- a/src/typechecker.cpp
+++ b/src/typechecker.cpp
@@ -7,20 +7,23 @@
 namespace TypeChecker {
 
 TypeChecker::TypeChecker() {
-	m_core.new_type_function(-1); // 0 | function
-	m_core.new_type_function(0);  // 1 | int
-	m_core.new_type_function(0);  // 2 | float
-	m_core.new_type_function(0);  // 3 | string
-	m_core.new_type_function(1);  // 4 | array
-	m_core.new_type_function(1);  // 5 | dictionary
-	m_core.new_type_function(0);  // 6 | boolean
-	m_core.new_type_function(0);  // 7 | unit
+	m_core.new_type_function(-1); // 0  | function
+	m_core.new_type_function(0);  // 1  | int
+	m_core.new_type_function(0);  // 2  | float
+	m_core.new_type_function(0);  // 3  | string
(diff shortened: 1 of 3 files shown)


src/typechecker.cpp
```diff
@@ -7,20 +7,23 @@
...
+	m_core.new_type_function(0);  // 1  | int
+	m_core.new_type_function(0);  // 2  | float
+	m_core.new_type_function(0);  // 3  | string
+	m_core.new_type_function(1);  // 4  | array
+	m_core.new_type_function(1);  // 5  | dictionary
+	m_core.new_type_function(0);  // 6  | boolean
+	m_core.new_type_function(0);  // 7  | unit
+	m_core.new_type_function(-1); // 8  | sum
+	m_core.new_type_function(-1); // 9  | product
+	m_core.new_type_function(-1); // 10 | record
```
SebastianMestre, line 20: This doesn't sound right to me.

A sum (or product, etc) type is a form of type function, it's not a particular type function.

I think `sum`, `product`, and `record` should be values in the `type_function_type` enum

SebastianMestre: To sum up: I think a type function should either be a var, a builtin, a sum, a product, or a record

SebastianMestre (changes requested):

SebastianMestre:
Also, unrelated, just to leave a written record: we really need to change the name of all the `foo_bar_type` enums to `FooBarType` or, even better, `FooBarTag`

src/typechecker.cpp
lushoBarlett: Yea perhaps this is a better way to do it

SebastianMestre (approved):
Yes

SebastianMestre:
makes sense

lushoBarlett:
> Also, unrelated, just to leave a written record: we really need to change the name of all the foo_bar_type enums to FooBarType or, even better, FooBarTag

I could definitely do this next. I would rather call them Tags when appropriate. But I don't know where would they be exactly so you might have to step in.

PR #127, opened 2020-09-13
Parse union typedefs

SebastianMestre:
I actually have some free time before the actual competition, so I decided to chip in.

I added some generally non-controversial code, and realized I didn't really know what to do next, because we have not decided on the syntax for type expressions.

I am not sure if you're familiar with BNF style syntax but here goes: this is what I think the syntax should be for types (roughly speaking).

BNF is just a way to define syntax kinda formally. You can put names on constructs with `::=`, `|` is an alternation, and a sequence of constructs just means concatenation. Plus, `...` just means "an arbitrary amount of repetitions of the stuff on the left".

```yacc
type_expr   ::= mono_type | type_func

mono_type   ::= type_var | type_term
type_var    ::= "@" identifier
type_term   ::= type_func "(<" mono_type "," ... ">)"

type_func   ::= union_expr | struct_expr | identifier

union_expr  ::= "union"  "{" declarator ";" ... ";" "}"
struct_expr ::= "struct" "{" declarator ";" ... ";" "}"

declarator  ::= identifier ":" mono_type
```

I did this because it really helps to remove ambiguity and make communication clearer (which is very helpful, because I am not very good at communicating clearly a lot of the time).

Let me know what you think about this particular syntax

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -118,6 +118,14 @@ struct DeclarationList : public AST {
 	    : AST {ast_type::DeclarationList} {}
 };
 
+// TODO: better name
+struct UnionTypeExpression : public AST {
+	std::vector<std::unique_ptr<AST>> m_declarations;
+
+	UnionTypeExpression()
+	    : AST {ast_type::UnionTypeExpression} {}
+};
+
 struct Declaration : public AST {
 	Token const* m_identifier_token;
 	std::unique_ptr<AST> m_type;  // can be nullptr
--- a/src/ast_type.hpp
+++ b/src/ast_type.hpp
@@ -26,6 +26,7 @@ constexpr const char* ast_type_string[] = {
 	"WhileStatement",
 
 	"TypeTerm",
+	"UnionTypeExpression",
 };
 
 enum class ast_type {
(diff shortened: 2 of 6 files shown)


lushoBarlett:
I'm comfortable with all of it, because it's been done or discussed. I can't really say anything about the union type syntax, because it _is_ consistent after all. But I will state that I find it a little uncomfortable to use. I'm biased obviously by the way Haskell does it.
```Haskell
data something = Blah | Bleh
data otherthing = (Blah, Bleh)
```
What I mean is: I think `union {  }` is appropriate for the language, but a little bloaty. Since bloatiness isn't particularly the language's concern so far (but my own), I have to agree that is should be like that.

SebastianMestre:
> What I mean is: I think `union { }` is appropriate for the language, but a little bloaty. Since bloatiness isn't particularly the language's concern so far (but my own), I have to agree that is should be like that.

I can't say I disagree, but I think developing a solid base is more important for now. We can add more syntactic sugar later on.

There are a bunch of things I want to try out when the language is a bit more mature, like decoupling concrete syntax from semantics even more. I think we could explore crazier/cooler syntactic options a lot more easily that way.

SebastianMestre:
Anyways, good talk. I don't think this code is very useful, so I'm closing this PR.

PR #128, opened 2020-09-15
Rename enums

lushoBarlett:
Currently I would just change
`something_type ===> SomethingType`
But I would like your opinion if any of them need a changing in the name whatsoever. Because I want to avoid extra work. I will list out every enum name we have as of the current master branch (with the changes done in this pr so far)
- ast_type -> ASTType
- token_type -> TokenType
- value_type
- exit_status_type
- BindingType
- mono_type
- kind_type
- type_function_type
- test_status


SebastianMestre:
I think I like `Tag` better as a suffix for this. This way, we don't have things with misleading names, like `MonoType`

lushoBarlett:
But for every single one?

SebastianMestre:
Yeah, no need to have two different conventions

lushoBarlett:
I'm too lazy to search and change all the variable and method names, if that's okay with you.

lushoBarlett:
I'm gonna rename the files, as well as the `something_string`s, what should both of them be named?

lushoBarlett:
Hmm I am quite fond of macros when dealing with code duplication, so I will use it

lushoBarlett:
So, it seems I can't do the macro trick with the token tags, because they differ... I'll leave it to you if you let me do it for them too (by changing symbols to words, or by doing it partially although that seems not that good of an idea)

SebastianMestre (approved):

lushoBarlett:
I'll take that as a "merge as is"

PR #131, opened 2020-09-16
Preparations for type declarations

lushoBarlett:
This implements type declarations. In reality I implement the existence of type functions and type variables. This way the type declaration will be possible for all the things we planned.


lushoBarlett:
I'm planning to have the type or kind be in a declaration node. Then the CTE can check and do whatever needs to be done.

SebastianMestre:
> I'm planning to have the type or kind be in a declaration node. Then the CTE can check and do whatever needs to be done.

Not really sure what you mean by this.

We do need a reified way to store type information in the compile time environment, which might be what you're getting at? Not too sure using Declaration for this is the right choice, though.

Oh well, you go ahead, try it out, see how it goes.

lushoBarlett:
Having a TypeDeclaration thingy has proven to be not that great in though experiments. How can the Parser tell if the thing on the right is a type? Can't be done without giving way too many tools that it should not be using. The only thing it can know for sure is "identifier" and then the typechecker knows if that's really correct.

SebastianMestre:
If you think about executing as code as having two evaluation stages (one at compile time, one at run time), and as expressions and types as being compile time values, you will find that declaring a type is not really a special case at all!

It just makes sense for a struct expression to be evaluated into a struct value at compile time, just like how integer expressions are evaluated into integer values at run time. (it is interesting to note that, under this mental model, most expressions evaluate to themselves in compile time)

lushoBarlett:
I'm thinking of ending it there, because more of this is just too much. It's better if I go slower. Mainly so you don't have to think too hard on too many lines of code

lushoBarlett:
I don't really know how to integrate parsing of the new stuff. I'm thinking of a `parse_type_declaration` that gets attempted in `parse_declaration`. If you agree on this, instead of reviewing and ending it here I might do that, else do you have any suggestions?

src/parser.cpp
```diff
@@ -1009,3 +1009,110 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_type_term() {
...
+			fields.push_back(std::move(field.m_result));
+			types.push_back(std::move(type.m_result));
+		}
+
+		auto s = std::make_unique<AST::Struct>();
+		s->m_fields = std::move(fields);
+		s->m_types = std::move(types);
+
+		return make_writer<std::unique_ptr<AST::AST>>(std::move(s));
+	}
```
SebastianMestre, line 1115: There is some pretty blatant duplication here (struct and union body is exactly the same) can you do something about it?

src/ast.hpp
```diff
@@ -225,6 +225,42 @@ struct TypeTerm : public AST {
 	    : AST {ASTTag::TypeTerm} {}
 };
 
+struct TypeVar : public AST {
+	Token const* m_token;
```
SebastianMestre, line 229: A comment about what this represents would be good

src/ast.hpp
```diff
@@ -225,6 +225,42 @@ struct TypeTerm : public AST {
...
+
+	std::string const& text() {
+		return m_token->m_text;
+	}
+
+	TypeVar()
+	    : AST {ASTTag::TypeVar} {}
+};
+
+struct Union : public AST {
```
SebastianMestre, line 239: This is not really an Union, but a union expression, or a union literal. Maybe change the name to reflect that?

SebastianMestre (changes requested):
I don't mind reviewing in bigger chunks. Actually, since I'm only doing it once a day, it would be better if the chunks were bigger.

lushoBarlett:
I don't really know how to integrate parsing of the new stuff. I'm thinking of a `parse_type_declaration` that gets attempted in `parse_declaration`. Do you have any suggestions?

SebastianMestre:
Here's what I've noticed: if we want `tuple{int,int}{3, 5}` to be valid code, we need to allow type functions in expressions.

Then, we need to restrict their usage on a later pass, like during type checking

Alternative, less general solution:

let a generalized expression be one of:  an expression, a type function, a type term, or a monotype var. When we parse a (top level?) declaration, the value should be a generalized expression

lushoBarlett:
> we need to allow type functions in expressions.

First of all, I agree. Second of all, I think I wanna leave that part to you actually (if you don't mind).

src/ast.hpp
```diff
@@ -226,7 +226,7 @@ struct TypeTerm : public AST {
 };
 
 struct TypeVar : public AST {
```
SebastianMestre, line 228: I meant a comment explaining what "typeVar" means

the comment on the next line is probably not necessary

src/parser.cpp
```diff
@@ -1010,7 +1010,47 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_type_term() {
 	return make_writer<std::unique_ptr<AST::AST>>(std::move(e));
 }
 
-Writer<std::unique_ptr<AST::AST>> Parser::parse_type_function() {
+Writer<std::unique_ptr<AST::AST>> Parser::parse_type_list(bool with_identifiers = false) {
```
SebastianMestre, line 1013: I know C++ is not very ergonomic for this kinda thing, but returning the two vectors (in a `std::pair`) would be nicer than creating a superfluous node

SebastianMestre:
```c++
Writer<std::pair<std::vector<std::unique_ptr<AST::AST>>, std::vector<std::unique_ptr<AST::AST>>>> Parser::parse_type_function() {}
```

lushoBarlett: I agree, this is exactly what I needed. But I just ended up creating another node :(

lushoBarlett:
So, that type is long AF. Maybe we could have a file with
```c++
template <typename T>
using Own<T> = std::unique_ptr<T>;
using OwnList<T> = std::vector<Own<T>>;
```
The second one is just for parser honestly but it would simplify the reading a lot. Definitely for another pr tho.

SebastianMestre (approved):
That type signature is quite hilarious

Looks great

PR #132, opened 2020-09-18
Type function metadata

lushoBarlett:


lushoBarlett:
I didn't forget, it's because I wanted to keep the diff short. Plus I wanted to go to sleep

lushoBarlett:
> Don't you think a hashtable from string to monotype would be more appropiate?

I do. This is just my way to get your feedback on stuff. I didn't have a clear idea of what needed to be done for this at all actually, so I just came up with something and made it into a draft. Now that you mention it, this is actually a nice idea, because name duplication would have resolved by now.

lushoBarlett:
It didn't fail on my machine. So it must have automatically included the string library. Huh.

SebastianMestre:
> ... Plus I wanted to go to sleep

very reasonable

src/typesystem.hpp
```diff
@@ -30,10 +26,10 @@ struct TypeFunctionData {
 
 // Concrete type function. If it's a built-in, we use argument_count
 // to tell how many arguments it takes. Else, for sum, product and record,
-// we store their structure.
+// we store their structure as a .
 struct TypeFunction {
 	int argument_count; // -1 means variadic
-	TypedAST::TypedAST* structure; // can be nullptr
+	std::unordered_map<std::string, MonoId> structure; // can be nullptr
 };
```
SebastianMestre, line 34: we will need a few more of these later, but this looks good for now

SebastianMestre:
I'm done with ioi now :)

adding a 'dummy' flag is probably a good idea

src/typesystem.cpp
```diff
@@ -37,7 +37,14 @@ MonoId TypeSystemCore::new_var() {
 
 MonoId TypeSystemCore::new_term(
     TypeFunctionId type_function, std::vector<int> args, char const* tag) {
-	int argument_count = type_function_data[type_function].argument_count;
+	TypeFunctionId tf_id = func_find(type_function);
+	TypeFunctionData& tf_data = type_function_data[tf_id];
+
+	// TODO: fail on dummy type function?
+	if (tf_data.type == TypeFunctionTag::Var)
+		assert(0 && "instantiating type function that could not be deduced");
```
SebastianMestre, line 45: Oh I hadn't noticed but, this check is wrong.

We should be able to create a term with a typefunc that isnt known. We just need to go back at the end and verify that we do eventually learn what it is.

SebastianMestre: Of course, in those cases we can't check that the argument is right. That also has to be done after type checking.

src/typesystem.hpp
```diff
@@ -30,9 +30,12 @@ struct TypeFunctionData {
 //
 // Dummy type functions are for unifying purposes only, but do not count
 // as 'deduced', because they were not created by the user
+
+using TypeFunctionStructure = std::unordered_map<std::string, MonoId>;
```
SebastianMestre, line 34: Hold off on the typedefs. They make code less readable

src/typesystem.hpp
```diff
@@ -36,7 +36,7 @@ using TypeFunctionStructure = std::unordered_map<std::string, MonoId>;
 struct TypeFunction {
 	int argument_count; // -1 means variadic
 	TypeFunctionStructure structure; // can be nullptr
-	bool is_dummy;
+	bool is_dummy = false;
```
SebastianMestre, line 39: We should really use brace notation for default values. It's just more idiomatic

 ```c++
bool is_dummy { false };
```

SebastianMestre (changes requested):

src/typesystem.cpp
SebastianMestre: The same thing goes for dummies. It should be possible to instanciate one, but we should make sure that it eventually gets resolved.

src/typesystem.hpp
SebastianMestre: This is not always the case, but when there is a different name for every possible instanciation of `std::unordered_map` and `std::vector`, it hinders more than it helps

SebastianMestre (approved):
I think we already discussed these points more than once, but, this is not what I had in mind

 - builtins, structs/unions, and tuples have different data, so they should be in separate arrays
 - also, when we are given a typefunc variable, or a dummy, we don't know what the argument count should be, so we shouldn't check it 

You don't seem to wanna do it that way, so I'll just approve this, and do it later.

lushoBarlett:
It's not that I don't wanna do it that way. I don't remember much of what we discussed about the implementation of this. The type system side is still new to me so it's natural than I don't wanna add many new elements at once.

With that said please tell me if you don't like something (or you'd rather it be done differently) I appreciate the guidance. Also if you think a pr should be closed, that's fine as well.

I'll merge it, since you approved it, and then you can adapt the parts that don't work.

PR #133, opened 2020-09-20
Record access

SebastianMestre:
Just implementing record access (`x.a`, where `x` is an expression of record type, and `a` is an identifier)

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -6,7 +6,15 @@
 
 namespace AST {
 
+namespace {
+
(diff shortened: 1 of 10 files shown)


SebastianMestre:
I ended up doing a bunch of style things

I mean, it had to be done eventually, so I'm ok with it.

SebastianMestre:
I would've liked to go all the way to typechecking, but that can be done later.

lushoBarlett (approved):
Looks cool, some cleanup always handy

PR #135, opened 2020-09-20
use # operator in Xmacros

SebastianMestre:
CPP interprets `"name" ` as a single token, not as `"` `name` `"`. To do this, CPP provides the `#` operator.


lushoBarlett (approved):
Did not know that, interesting

PR #138, opened 2020-09-21
Typed ast enum

### LINKED ISSUE #134: Noisy warnings
Currently, we use the same `ASTTag` to list out all the possible variants in both AST and TypedAST. This is starting to be a bit troublesome because the two are diverging, meaning that whenever we `switch` over a TypedAST's `type()`, we get a bunch of noisy warnings.

Of course, one could add a default case, but that is just treating the symptom. Furthermore, it silences the warning, opening up a way for bugs.

an example:

We don't need `TypeTerm`, `UnionExpression`, `StructExpression`, etc. `TypedAST` nodes. Instead, we need a node that holds a type function value, and a node that holds a mono type value (i.e. an int that is an index into one of the type system core arrays). This hypothetical node is definitely not needed in `AST`, further widening the gap.

E:

There is also the obvious example that we don't have `TypedAST::BinaryExpression`

SebastianMestre:
Closes #134 

I added a brand new TypedASTTag enum, which lacks some of the values in ASTTag.

### DIFF
--- a/src/interpreter/eval.cpp
+++ b/src/interpreter/eval.cpp
@@ -13,7 +13,7 @@ namespace Interpreter {
 
 gc_ptr<Value> eval(TypedAST::DeclarationList* ast, Environment& e) {
 	for (auto& decl : ast->m_declarations) {
-		assert(decl->type() == ASTTag::Declaration);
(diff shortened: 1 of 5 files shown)


lushoBarlett (approved):
Looks good

PR #140, opened 2020-09-22
Typecheck record access

SebastianMestre:
I don't think this is fine. I get the feeling I missed a check in func_unify, but I don't know what it is.

### DIFF
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -53,8 +53,7 @@ void typecheck(TypedAST::Declaration* ast, Frontend::CompileTimeEnvironment& env
 	std::cerr << "Typechecking " << ast->identifier_text() << '\n';
 #endif
 
-	// we use a hidden typevar to get it to generalize if needed
-	ast->m_value_type = env.new_hidden_type_var();
+	ast->m_value_type = env.new_type_var();
 	env.declare(ast->identifier_text(), ast);
 
 	// this is where we implement rec-polymorphism.
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -225,6 +225,17 @@ void typecheck(TypedAST::TernaryExpression* ast, Frontend::CompileTimeEnvironmen
 	ast->m_value_type = ast->m_then_expr->m_value_type;
 }
 
+void typecheck(TypedAST::RecordAccessExpression* ast, Frontend::CompileTimeEnvironment& env) {
+
+	// should this be a hidden type var?
+	MonoId member_type = env.new_type_var();
+	TypeFunctionId dummy_tf = env.m_typechecker.m_core.new_dummy_type_function(
+	    TypeFunctionTag::Record, {{ast->m_member->text(), member_type}});
+	MonoId term_type = env.m_typechecker.m_core.new_term(dummy_tf, {});
+
+	ast->m_value_type = term_type;
+}
+
 void typecheck(TypedAST::DeclarationList* ast, Frontend::CompileTimeEnvironment& env) {
 
 	// two way mapping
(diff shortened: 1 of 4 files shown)


SebastianMestre:
I think it's good now.

It's not pretty, though

src/typesystem.hpp
```diff
@@ -17,9 +17,9 @@ enum class TypeFunctionTag { Var, Builtin, Sum, Product, Record };
 // For the purposes of the type system, we only care about the amount
 // of argument it takes.
 
-// If it's a var, equals points to another TypeFunctionData.
+// If it's a var, equals points to another TypeFunctionHeader.
 // Else it points to a TypeFunction.
-struct TypeFunctionData {
+struct TypeFunctionHeader {
```
lushoBarlett, line 22: I'm all for this name. I really like it.

src/typesystem.cpp
```diff
@@ -208,17 +208,69 @@ void TypeSystemCore::func_unify(TypeFunctionId a, TypeFunctionId b) {
...
-		b_header.equals = a;
-	else
-		assert(0 and "unifying two different known type functions");
-}
+	// ensure a is a var if at least one of them is a var
+	if (type_function_header[b].type == TypeFunctionTag::Var)
+		std::swap(a, b);
+
+	if (type_function_header[a].type == TypeFunctionTag::Var) {
+		type_function_header[a].equals = b;
```
lushoBarlett, line 220: I like this style, I'll be sure to keep `std::swap` in mind for the future.

src/typesystem.cpp
```diff
@@ -208,17 +208,69 @@ void TypeSystemCore::func_unify(TypeFunctionId a, TypeFunctionId b) {
...
+			// this way, if more fields get added to a or b, both get updated
+			//
+			// TODO: think more thoroughly about this...
+			// I get the feeling that we shouldn't really do this if b is a
+			// polymorphic record. why? Because b will have type variables that
+			// are bound to a forall kinda thing, and thus should NOT be unified.
+			// instead, we should do something similar to inst_fresh, but for
+			// type funcs.
+			type_function_header[a].type = TypeFunctionTag::Var;
+			type_function_header[a].equals = b;
```
lushoBarlett, line 270: What do you mean?

SebastianMestre: if we have a dummy-record in the type of a top-level declaration, users of that declaration can unify stuff with it and modify it.

This is not ideal if we want to have polymorphism on records.

src/typecheck.cpp
```diff
@@ -226,14 +226,18 @@ void typecheck(TypedAST::TernaryExpression* ast, Frontend::CompileTimeEnvironmen
 }
 
 void typecheck(TypedAST::RecordAccessExpression* ast, Frontend::CompileTimeEnvironment& env) {
+	typecheck(ast->m_record.get(), env);
```
lushoBarlett, line 231: Is this function defined somewhere?

lushoBarlett (approved):
I think this is correct.

I'm not sure if a test is possible for this yet. If it is, please consider adding it.

src/typecheck.cpp
SebastianMestre: `m_record.get()` is a `TypedAST*`.

`typecheck(TypedAST*, CompileTimeEnvironment&)` is the very last definition in this file.

src/typesystem.cpp
SebastianMestre: Essentially, dummy records don't get generalized, so they can be mutated after we are done typechecking the declaration they came from.

SebastianMestre:
> I think this is correct.
> 
> I'm not sure if a test is possible for this yet. If it is, please consider adding it.

Uhh... kinda? I'll add a test, you let me know if it actually helps at all.

PR #142, opened 2020-09-24
parse record access expressions directly

SebastianMestre:
This makes the representation of record access nodes a bit nicer to work with


lushoBarlett (approved):
I think it's less bloaty this way so I'm for it

PR #144, opened 2020-09-24
Cleanup convert ast

### LINKED ISSUE #137: convertAST family of functions should return unique_ptr
The whole `get_unique` deal is just nonsense. A proper modern C++ API should be safe from the get go.

If you don't want a safe owning pointer, you can just `release()` it.

SebastianMestre:
Closes #137 

I made it so that convertAST returns a unique ptr directly, and renamed it to convert_ast, to better fit the naming convention.

Also, I got rid of conversion for object literals, as we don't really know how we want to implement that, yet.


lushoBarlett (approved):
One less thing to worry about

PR #146, opened 2020-09-24
Typechecker refactor

### LINKED ISSUE #145: The architecture of the type checker is all wrong
As it stands, the tc is not nice to work with. This is because it was built with a mental model that is completely backwards!

Now that we have a better idea of what things need it needs to do, we have a good opportunity to re-architect it.

### The problems

The biggest mistake is that the type checker is a member of the environment, and not the other way around. This is clear from how often we say `env.m_typechecker` in `typecheck`.

Another mistake is how the `TypeSystemCore` has too many responsibilities. For instance, `generalize` should be in the `TypeChecker`, instead (It takes the environment as an argument! wtf?!) . The core should only be responsible for unification.

Another thing is how the scope hierarchy is not stored explicitly. I feel like if we actually stored the scope tree, and a pointer to its surrounding scope on every `TypedAST` node, a bunch of things would be simplified.

### The proposal

Here goes a general mock-up of how things would look.

Another thing that needs addressing is how type functions can reference other type functions that come *after* them in the source code. I don't have a definitive solution to this yet, but I think the new architecture will make this easier to do.

```c++
struct TypedAST {
  TypedASTTag m_tag;
  MonoId m_value_type;
  Scope* m_surrounding_scope;
};

struct TypeChecker {
  TypeSystemCore m_core;
  CompileTimeEnvironment m_env;

  PolyId generalize(MonoId);
};

struct CompileTimeEnvironment {
  bucket_array<Scope> m_scopes;
  Scope m_global_scope;
  Scope* m_current_scope;
};

struct Scope {
  std::set<MonoId> m_mono_vars;
  std::map<std::string, Declaration> m_bindings;
};

struct TypeSystemCore {
  std::vector<MonoHeader> mono_header;
  // etc ...

  MonoId find(MonoId);
  void unify(MonoId, MonoId);

  TypeFuncId find_func(TypeFuncId);
  void unify_func(TypeFuncId, TypeFuncId);
};
```

There's probably more things that I haven't thought about. Thoughts?

SebastianMestre:
These are some of the changes mentioned in #145.

They aren't tricky changes, so I figured just trying it out would be good, even if we don't keep them

EDIT: closes #145

### DIFF
--- a/src/typecheck.hpp
+++ b/src/typecheck.hpp
@@ -4,11 +4,9 @@ namespace TypedAST {
 struct TypedAST;
 }
 
-namespace Frontend {
-struct CompileTimeEnvironment;
-}
-
 namespace TypeChecker {
+
+struct TypeChecker;
 
 /*
  * Runs type deduction and checking on the given ast.
@@ -19,5 +17,5 @@ namespace TypeChecker {
  * PRECONDITION: match_identifiers has already been called
  * the given ast.
  */
-void typecheck(TypedAST::TypedAST* ast, Frontend::CompileTimeEnvironment&);
+void typecheck(TypedAST::TypedAST* ast, TypeChecker&);
 } // namespace TypeChecker
--- a/src/typechecker.cpp
+++ b/src/typechecker.cpp
@@ -21,6 +21,115 @@ TypeChecker::TypeChecker() {
 	m_core.new_term(3, {}, "builtin string"); // 2 | string(<>)
 	m_core.new_term(6, {}, "builtin bool");   // 3 | boolean(<>)
 	m_core.new_term(7, {}, "builtin unit");   // 4 | unit(<>)
+
+	// TODO: put this in a better place
+	// TODO: refactor, figure out a nice way to build types
+	// HACK: this is an ugly hack. bear with me...
+
+	{
+		auto var = new_var();
+		auto poly = m_core.new_poly(var, {var});
+		m_env.declare_builtin("print", poly);
+	}
+
+	{
+		auto var_mono_id = new_var();
+		auto var_id = m_core.mono_data[var_mono_id].data_id;
+
+		auto array_mono_id = m_core.new_term(
+		    BuiltinType::Array, {var_mono_id}, "array");
+
+		{
+			auto term_mono_id = m_core.new_term(
+			    BuiltinType::Function,
+			    {array_mono_id, var_mono_id, mono_unit()},
+			    "[builtin] (array(<a>), a) -> unit");
+
+			auto poly_id = m_core.new_poly(term_mono_id, {var_id});
+			m_env.declare_builtin("array_append", poly_id);
+		}
+		{
+			auto term_mono_id = m_core.new_term(
+			    BuiltinType::Function,
+			    {array_mono_id, mono_int()},
+			    "[builtin] (array(<a>)) -> int");
+
+			auto poly_id = m_core.new_poly(term_mono_id, {var_id});
+			m_env.declare_builtin("size", poly_id);
+		}
+		{
+			auto term_mono_id = m_core.new_term(
+			    BuiltinType::Function,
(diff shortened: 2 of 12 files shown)


src/typecheck.hpp
```diff
@@ -4,12 +4,10 @@ namespace TypedAST {
...
 }
 
-namespace Frontend {
-struct CompileTimeEnvironment;
-}
-
 namespace TypeChecker {
 
+struct TypeChecker;
+
```
lushoBarlett, line 10: Also a great deal of sense, since we are _typechecking_, to use the _TypeChecker_

lushoBarlett (commented):
It's a good reordering of things in my opinion. I was thinking if maybe `has_type_var` should be in the CTE, perhaps it's better to move it to the TC as well?

SebastianMestre:
> It's a good reordering of things in my opinion. I was thinking if maybe `has_type_var` should be in the CTE, perhaps it's better to move it to the TC as well?

Yeah, maybe. I didn't do that because the implementation of `has_type_var` only uses the environment

src/typechecker.cpp
```diff
@@ -21,6 +21,115 @@ TypeChecker::TypeChecker() {
...
+		auto term_mono_id = m_core.new_term(
+		    BuiltinType::Function,
+		    {var_mono_id, var_mono_id, mono_boolean()},
+		    "[builtin] (a, a) -> Bool");
+
+		auto poly_id = m_core.new_poly(term_mono_id, {var_id});
+
+		m_env.declare_builtin("<", poly_id);
+		m_env.declare_builtin("==", poly_id);
+	}
```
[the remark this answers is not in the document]
SebastianMestre: yeah, the weird architectural choices I made ended up forcing even more weird choices

src/typecheck.hpp
SebastianMestre: When you change an abstraction and realize that things suddenly make sense: it's just so... nice.

SebastianMestre:
> It's a good reordering of things in my opinion. I was thinking if maybe `has_type_var` should be in the CTE, perhaps it's better to move it to the TC as well?

I looked at it, and it pretty much works on the internals of CTE. The best way to put in TC would be to just add a method that calls the one that already exists. I don't see much value in that, so maybe we can delay it to a later pr?

Also, the idea that we ask the CTE if it has a typevar kinda appeals to me.

lushoBarlett:
> Also, the idea that we ask the CTE if it has a typevar kinda appeals to me.

I guess that makes sense

lushoBarlett (commented):
Wait, there's no need for Binding anymore? Are argument declarations and regular declarations no different now? I can't tell why they had to be different in the first place

SebastianMestre:
> Wait, there's no need for Binding anymore? Are argument declarations and regular declarations no different now? I can't tell why they had to be different in the first place

The reason is that we didn't allow monomorphic declarations before

SebastianMestre:
Also, can you approve or reject instead of just commenting?

lushoBarlett:
I comment because I need insight in your decisions

lushoBarlett (approved):

SebastianMestre:
> I comment because I need insight in your decisions

Comment using the comment box at the bottom of the page, not the review feature. This way, I don't have to re-request a review after every comment.

lushoBarlett:
Ah so that's what you mean. Yea I can do that

SebastianMestre:
> Ah so that's what you mean. Yea I can do that

Ok. So... I though you were submitting comments as reviews on purpose. And I really didn't like that because it makes what I'm supposed to do unclear.

But it looks like it's just Github being bad about how this is handled.

Sorry, I was a bit mean, unjustly.

lushoBarlett:
> I though you were submitting comments as reviews on purpose.

Don't apologize, you were partially right actually. I do use the "comment" review sometimes. I will stop doing that tho.

PR #147, opened 2020-09-25
Store typefuncs

SebastianMestre:
Some preliminary work before we start evaluating typefuncs

EDIT:

Plans: implement the notion of metatypes.

this lets us do some rudimentary 'typechecking' that informs how to handle a compile time value during actual typechecking.

Why? typecheck needs to do different things depending on whether an expression is a monomorphic value, a polymorphic value or a typefunc.

### DIFF
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -4,7 +4,7 @@
 #include "tarjan_solver.hpp"
 #include "typechecker.hpp"
 #include "typed_ast.hpp"
-#include "typesystem_types.hpp"
+#include "typechecker_types.hpp"
 
 #include <cassert>
 #include <iostream>
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -86,7 +86,6 @@ void typecheck(TypedAST::Declaration* ast, TypeChecker& tc) {
 		std::cerr << "@@ Has " << poly_data.vars.size() << " variables\n";
 		std::cerr << "@@ It is equal to:\n";
 		tc.m_core.print_type(poly_data.base);
-
 	}
 #endif
 }
@@ -104,9 +103,11 @@ void typecheck(TypedAST::Identifier* ast, TypeChecker& tc) {
 		else
 			ast->m_value_type = declaration->m_value_type;
 	} else if(meta_type == tc.meta_typefunc()){
+		assert(0 && "Accessed a typefunc by name (not implemented)");
 		// we are a type function.
 		// TODO: not too sure what needs to be done...
 	} else {
+		assert(0 && "Accessed a name with unknown metatype (not implemented)");
 		// meta type var
 		// TODO: not too sure what needs to be done...
 	}
@@ -165,6 +166,7 @@ void typecheck(TypedAST::FunctionLiteral* ast, TypeChecker& tc) {
 			int mono = tc.new_var();
 
 			arg_types.push_back(mono);
+			arg_decl.m_meta_type = tc.meta_value();
 			arg_decl.m_value_type = mono;
 		}
 
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -17,32 +17,26 @@ namespace TypeChecker {
 // At least, the alternative just sucks.
 
 void typecheck(TypedAST::NumberLiteral* ast, TypeChecker& tc) {
-	ast->m_meta_type = tc.meta_value();
 	ast->m_value_type = tc.mono_float();
 }
 
 void typecheck(TypedAST::IntegerLiteral* ast, TypeChecker& tc) {
-	ast->m_meta_type = tc.meta_value();
 	ast->m_value_type = tc.mono_int();
 }
 
 void typecheck(TypedAST::StringLiteral* ast, TypeChecker& tc) {
-	ast->m_meta_type = tc.meta_value();
 	ast->m_value_type = tc.mono_string();
 }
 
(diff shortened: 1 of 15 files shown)


SebastianMestre:
I don't know how to implement this yet. I already threw away my first attempt, and I am willing to throw this one away, too.

SebastianMestre:
It tried doing this with just an enum, but that way you can't express that two things have the same metatype but you don't know what it is. (yeah, it's the same problem we had when we started implementing the typechecker)

I went with a unification based solution. I put it in a separate structure, but maybe that's not the way to go. There are still many things to figure out.

lushoBarlett:
So, on first glance I believe this is just a higher level TSC, which is fine. Is this an equivalent to the `TypeVar` stuff over at the TSC that I did? Because this is a better alternative than having it all mushed there.

Also I was thinking that, since this is the third thing with this structure, maybe we can abstract it into a reusable module. I really get the vibes that it can be done so I'm gonna try it sometime this weekend and see how it goes

SebastianMestre:
Thinking about it, there is a bit of a hierarchy of type-like things:

 - metatype is:
   - value has:
     - monotype
     - polytype
   - typefunc is:
     - builtin
     - user-defined

I can't think of a particularly good way to model this, so I'm just going to start hacking and see what comes out

SebastianMestre:
> So, on first glance I believe this is just a higher level TSC, which is fine. Is this an equivalent to the `TypeVar` stuff over at the TSC that I did? Because this is a better alternative than having it all mushed there.

Is it trying to do the same thing... Maybe? I don't really know. I haven't thought this through, tbh. If I had to say, though: yes, it is trying to do the same thing.

> Also I was thinking that, since this is the third thing with this structure, maybe we can abstract it into a reusable module. I really get the vibes that it can be done so I'm gonna try it sometime this weekend and see how it goes.
> 
> E: I mean the whole union find header var stuff

It can probably be done. If we take the one that's used for unifying monotypes, we can probably make it more generic and use it here, too.

This one is simpler, though: the 'terms' don't have arguments. It's essentially a normal union-find where some of the nodes can't be unified.

SebastianMestre:
I'm starting to get the feeling that we should do metatype checking before we even start doing typechecking

It might be possible to do it in a single pass, but I am getting really confused about it.

SebastianMestre:
I know, it's long. There was (and still is) a lot to do

src/typechecker.cpp
```diff
@@ -192,4 +200,12 @@ MonoId TypeChecker::mono_unit() {
 	return 4;
 }
 
+MetaTypeId TypeChecker::meta_value() {
+	return 0;
+}
```
lushoBarlett, line 205: In the future it might be useful to differentiate mono_values from poly_values (basically just functions). I will go a step further and say this will be useful for the idea of type_mappings (type functions that are not constructors, but proper functions) which you suggested in the code somewhere and I like that idea.

src/typecheck.cpp
```diff
@@ -87,17 +86,24 @@ void typecheck(TypedAST::Identifier* ast, TypeChecker& tc) {
...
-		mono = tc.m_core.inst_fresh(declaration->m_decl_type);
+	MetaTypeId meta_type = tc.m_meta_core.find(declaration->m_meta_type);
+
+	ast->m_meta_type = meta_type;
+	if (meta_type == tc.meta_value()) {
+		// here we implement the [var] rule
+		if (declaration->m_is_polymorphic)
+			ast->m_value_type = tc.m_core.inst_fresh(declaration->m_decl_type);
+		else
+			ast->m_value_type = declaration->m_value_type;
```
lushoBarlett, line 97: Perfect ternary operator here

src/typecheck.cpp
```diff
@@ -87,17 +86,24 @@ void typecheck(TypedAST::Identifier* ast, TypeChecker& tc) {
...
+	} else if(meta_type == tc.meta_typefunc()){
```
lushoBarlett, line 98: Insert a space before open brace

lushoBarlett (approved):
So, yea, this is a cool concept, maybe implemented poorly? but that's the codebase's fault at this point.

What I mean is that there's a bunch of places where improvements / replacements need to be made. But it's good.

I'll approve it, obviously.

src/typechecker.cpp
lushoBarlett: Differentiating monos from polys is good for removing a second flag later (is_polimorphic), which appeals greatly to me. Plus it makes sense to me to have all possible type differentiations centralized in the meta type system.

SebastianMestre: I had that same thought, but I ran into some issues:

When you access a poly_value, you always instianciate it first, which turns it into a mono_value. This means that just because you assign an expression into another, the meta type could change, and that does not play nicely with unification.

I "fixed" this by having mono values and poly values be the same thing at the meta type level

src/typecheck.cpp
SebastianMestre: good point

src/typecheck.cpp
SebastianMestre: will do

lushoBarlett: Oh and also space before open parenthesis, completely did not see that

src/typechecker.cpp
lushoBarlett: Ah I see, well that's a mild bummer.

PR #149, opened 2020-09-26
Union-find module

lushoBarlett:
Not sure if this is a good idea, but it is an idea.

I believe the whole deal with the type system being a bunch of classes is not actually relevant (speed-wise), so I took an approach which uses a templated base class with some default behavior. It looks good and useful to me, but you might disagree.

The whole point here is that you can do stuff like
```c++
enum class MyTagSystem { ... };
struct MyHeader { ... };
struct MyData { ... };

struct MyStructure : Unification::Core<MyData, MyHeader> {
    // override whatever I need
};
```
This way the interface is the same everywhere, so for example at the TSC you could have
```c++
struct TypeSystemCore {
    MetaTypes metas    
    TypeFunctions tfs;
    MonoTypes monos;
    PolyTypes polys;
    TypeVars vars;
};
```
each with their own unique functionality but the same `something.header` or `something.data` vectors, or the same `something.new_var()` method.

In this pr I will also replace the current union find implementations and see how it looks.


lushoBarlett:
So, I was trying to extract the functionality of type functions into a struct to use the Unification::Core. But it turns out it breaks (heap buffer overflow) and I don't know why. Anyway I feel like this was maybe a bad idea to I'm gonna just close it here.

If you think there's potential, reopen.

SebastianMestre:
This is probably a thing that needs to be done. This is not how I'd do it, though.

SebastianMestre:
The last one was a very unhelpful comment. Here is how I would go about this:

https://gist.github.com/SebastianMestre/71a9c9691d5e4437d265051405ba94e

My implementation has a bunch of issues, but I think the general idea is the way to go.

lushoBarlett:
It seems I don't have permissions to see that?

PR #150, opened 2020-09-28
add a TypedAST node that holds a type func ID

SebastianMestre:
An idea: compile time is just another moment for evaluation, just that most expressions evaluate to themselves, and some evaluate to monotypes or type functions.

Example: `struct { x : int }` evaluates into a type function

### DIFF
--- a/src/typechecker.cpp
+++ b/src/typechecker.cpp
@@ -176,8 +176,16 @@ void TypeChecker::declare_builtin(std::string const& name, MetaTypeId meta_type,
 	m_builtin_declarations.push_back({});
 	TypedAST::Declaration* decl = &m_builtin_declarations.back();
 	decl->m_meta_type = meta_type;
-	decl->m_decl_type = poly_type;
-	decl->m_is_polymorphic = true;
+	// BIG HACK:
+	// if we are declaring a typefunc, 'poly_type' is actually a TypeFunctionId
+	if (meta_type == meta_typefunc()) {
+		auto handle = std::make_unique<TypedAST::TypeFunctionHandle>();
+		handle->m_value = poly_type;
+		decl->m_value = std::move(handle);
+	} else {
+		decl->m_decl_type = poly_type;
+		decl->m_is_polymorphic = true;
+	}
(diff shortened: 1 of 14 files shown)


SebastianMestre:
whew... this was exhausting.

SebastianMestre:
RATIONALE:

It makes sense to think about a compile-time declarative language that deals in types and expressions. Using this language, we build up the type functions and monotypes that are used in the final program.

In the implementation of this language, TypedAST::TypedAST is analogous to Interpreter::Value.

lushoBarlett:
Makes sense, although I will read this tomorrow because its suicide to do it now.

src/typechecker.cpp
```diff
@@ -176,8 +176,16 @@ void TypeChecker::declare_builtin(std::string const& name, MetaTypeId meta_type,
...
+	// BIG HACK:
+	// if we are declaring a typefunc, 'poly_type' is actually a TypeFunctionId
+	if (meta_type == meta_typefunc()) {
+		auto handle = std::make_unique<TypedAST::TypeFunctionHandle>();
+		handle->m_value = poly_type;
+		decl->m_value = std::move(handle);
+	} else {
+		decl->m_decl_type = poly_type;
+		decl->m_is_polymorphic = true;
+	}
```
lushoBarlett, line 188: Not a fan of this, will look into ways of cleaning this up

SebastianMestre: yeah it's disgusting

lushoBarlett (approved):
So, yea, most of the code added was just tree traversal boilerplate, and it only does one thing for now. The TypeFunctionHandle is a straightforward approach, similar to how identifier points to the declaration.

I must say that I will attempt to read all of the code and look for ways to reduce the amount of lines in it. I get very confused and it takes way too long for me to review this stuff.

I think this is correct.

SebastianMestre:
The tricky bits are in `type_func_from_ast`, `ct_eval(DeclarationList)`, and, as you noted, `Typechecker::declare_builtin`

PR #151, opened 2020-09-28
Parser cleanup

SebastianMestre:
I wanted to do some work on the type system, but the weak-typedness of the AST is getting too unbearable to work with for me.

I originally did it this way to avoid having to think about it too much while the structure of the ast was constantly changing. I feel like it has stabilized enough that it's no longer an advantage. Rather, having to explicitly check types that should be known at compile time 'just in case' has gotten too annoying to keep bearing with.

There are a lot of parts of the AST whose types can be specified in more detail (`DeclarationList::m_declarations` is an obvious example), but I think this is a good start.

I chose this node because it's the part that I am working on integrating with the type systems, and it makes some code simpler to have explicit types

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -554,8 +554,8 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_ternary_expression() {
 	return make_writer<std::unique_ptr<AST::AST>>(std::move(e));
 }
 
-Writer<std::unique_ptr<AST::AST>> Parser::parse_identifier() {
-	Writer<std::unique_ptr<AST::AST>> result = {
+Writer<std::unique_ptr<AST::Identifier>> Parser::parse_identifier() {
+	Writer<std::unique_ptr<AST::Identifier>> result = {
 	    {"Parse Error: Failed to parse identifier"}};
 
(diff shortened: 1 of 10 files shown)


SebastianMestre:
There are tons of redundant indirections in the ast, I'll try to reduce some of them.

SebastianMestre:
There's still more, but I feel good leaving it here for now.

src/parser.cpp
```diff
@@ -942,7 +942,7 @@ Writer<std::unique_ptr<AST::AST>> Parser::parse_statement() {
 				return result;
 			}
 
-			return declaration;
+			return make_writer<std::unique_ptr<AST::AST>>(std::unique_ptr<AST::AST>(declaration.m_result.release()));
```
lushoBarlett, line 945: Shouldn't you call `make_unique` here?

lushoBarlett (approved):
Removing indirections is nice

src/parser.cpp
SebastianMestre: No. make_unique is for creating new objects. Here, we steal a pointer from an already-existing unique_ptr

PR #153, opened 2020-09-29
Unification Engine

lushoBarlett:
Implements a common union find algorithm found several times on the codebase. I used Seba's implementation, that inyects a generic behavior in the unification function.


lushoBarlett:
I had to add some things to the Unification::Core for this one to work, and also be nicer to work _with_.

I added another auxiliary finder, it gives you the function of a term. And I also needed to change the unify_function interface because type function unification is weird (because of dummies). I believe this ain't that bad, given that I provide auxiliary finders to shorten code.

I also want to note that making the Unification::Core hold the vector in which the custom "extended" data of a term (which was a thing that I think you mentioned) does not seem good, but I will look into it further down the line of this pr.

src/unification.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+	enum class Tag { Var, Term, };
+
+	struct NodeHeader {
+		Tag tag;
+		int data_idx;
+	};
+
+	struct TermData {
+		int function_id; // possibly an external id
```
SebastianMestre, line 18: More like "an external id"

src/unification.hpp
```diff
@@ -0,0 +1,36 @@
...
+		std::vector<int> argument_idx;
+	};
+
+	std::function<void(int,int)> unify_function;
```
SebastianMestre, line 22: This doesn't take function IDs, but node IDs, right?

SebastianMestre: Just trying to understand what you changed

src/unification.cpp
```diff
@@ -0,0 +1,82 @@
...
+bool Core::occurs(int v, int i){
+	assert(node_header[v].tag == Tag::Var);
+	assert(node_header[v].data_idx == v);
+
+	i = find(i);
+
+	if (i == v)
+		return true;
+
+	assert(node_header[i].tag == Tag::Term);
```
SebastianMestre, line 16: did I write this? hahah. I don't think that assertion is right.

SebastianMestre: `i` can still be a Var that's not equal to `v`, can't it?

src/unification.hpp
lushoBarlett: Realized that later

src/unification.hpp
lushoBarlett: Yes, it now takes "header references"

src/typesystem.hpp
```diff
@@ -6,32 +6,23 @@
...
 // Dummy type functions are for unifying purposes only, but do not count
-// as 'deduced', because they were not created by the user
-
+// as 'deduced', because they were not created by the user/
+//
+// TODO: change for polymorphic approach
 struct TypeFunctionData {
+	TypeFunctionTag tag;
 	int argument_count; // -1 means variadic
 	std::unordered_map<std::string, MonoId> structure; // can be nullptr
```
SebastianMestre, line 27: lol, did I write that `// can be nullptr` comment? I wonder what I was thinking?

src/unification.cpp
lushoBarlett: It shouldn't, the way we use it, in any case we should return false instead of asserting

src/typesystem.cpp
```diff
@@ -6,6 +6,56 @@
...
+			// this way, if more fields get added to a or b, both get updated
+			//
+			// TODO: think more thoroughly about this...
+			// I get the feeling that we shouldn't really do this if b is a
+			// polymorphic record. why? Because b will have type variables that
+			// are bound to a forall kinda thing, and thus should NOT be unified.
+			// instead, we should do something similar to inst_fresh, but for
+			// type funcs.
+			//
+			// NOTE: this makes the unify_function need to access node_header
```
SebastianMestre, line 47: Yeah, that makes sense.

src/typesystem.hpp
lushoBarlett: No, that was me actually... I realize now that I meant it can be _empty_ (when unused)

src/typesystem.cpp
```diff
@@ -66,30 +114,20 @@ PolyId TypeSystemCore::new_poly(MonoId mono, std::vector<MonoId> vars) {
...
 TypeFunctionId TypeSystemCore::new_dummy_type_function
     (TypeFunctionTag type, std::unordered_map<std::string, MonoId> structure) {
-	TypeFunctionId type_function_var = type_function_header.size();
-
-	type_function_header.push_back({type, type_function_data.size()});
-	type_function_data.push_back({-1, structure, true});
-
-	return type_function_var;
+	TypeFunctionId id = m_tf_core.new_term(m_type_functions.size(), {});
+	m_type_functions.push_back({type, -1, structure, true});
```
SebastianMestre, line 127: Somewhat unrelated, but maybe 0 is better than -1 here? (I mean the argument count)

SebastianMestre:
Ignore this comment

src/typesystem.cpp
lushoBarlett: Probably. 0 is more "safe" I think

lushoBarlett:
> Ignore this comment
> 
> OK Github is really dumb.

I think it's okay that GitHub expects a comment when you review with a comment, perhaps the right approach was not to review (were you forced to do so?)

src/typesystem.cpp
```diff
@@ -189,92 +227,6 @@ void TypeSystemCore::unify(MonoId a, MonoId b) {
...
-			type_function_header[a].type = TypeFunctionTag::Var;
-			type_function_header[a].equals = b;
-
-		} else {
-			// neither one is a dummy, and they are different, so we can't unify.
-			assert(0 and "unifying two different known type functions");
-		}
-	}
-}
-
```
SebastianMestre, line 276: good riddance

SebastianMestre:
> > Ignore this comment
> > OK Github is really dumb.
> 
> I think it's okay that GitHub expects a comment when you review with a comment, perhaps the right approach was not to review (were you forced to do so?)
> 
> E: I now read the rest. Github is dumb

Yeah. I might start to leave regular comments with links to lines from now on. It's a lot more work, but the other thing is just wrong, and I am a bit particular about these things.

SebastianMestre:
> ... I also needed to change the unify_function interface because type function unification is weird ...

Yup, I looked at the change, it makes a lot of sense.

> I also want to note that making the Unification::Core hold the vector in which the custom "extended" data of a term (which was a thing that I think you mentioned) does not seem good, but I will look into it further down the line of this pr.

I agree. Putting that stuff in its own container sounds way nicer.

SebastianMestre:
test comment. This change is very nice!

https://github.com/SebastianMestre/Jasper/blob/d02b8dee444d50d05c68b2b58d4275c61e54e201/src/ct_eval.cpp#L123-L125

lushoBarlett:
My method for reviewing is:
- Read commit by commit to understand the whole process
- Read the full diff and pin down the interesting bits (also make sure you don't comment on something that was later changed)
- Go back to commits and leave comments

But I'm now doubtful, I think I remember leaving regular comments while at the full diff page?

src/unification.cpp
```diff
@@ -52,7 +52,7 @@ void Core::unify(int i, int j) {
 		int vi = node_header[i].data_idx;
 		int vj = node_header[j].data_idx;
 
-		unify_function(term_data[vi].function_id, term_data[vj].function_id);
+		unify_function(i, j);
```
SebastianMestre: Test comment 2

lushoBarlett: _It reviewed again. Are you kidding me??_

src/typesystem.cpp
```diff
@@ -6,6 +6,56 @@
...
+		if (b_data.is_dummy) {
+			std::swap(a, b);
+			std::swap(a_data, b_data);
+		}
+
+		if (a_data.is_dummy) {
+			for (auto& kv_a : a_data.structure) {
+				auto kv_b = b_data.structure.find(kv_a.first);
+
+				if (kv_b == b_data.structure.end())
```
lushoBarlett, line 26: Leaving single comment on random line

lushoBarlett: Okay. This *has* to be non-intentional. There's clearly a "single comment" and a "review" button. It must be a bug.

SebastianMestre:
In this page

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request

it says:

> You can also comment on specific sections of a file on a pull request's Files changed tab in the form of individual line comments or as part of a pull request review.

Which makes it seem like line comments are not reviews.

But then, it says

> Adding line comments is a great way to discuss questions about implementation or provide feedback to the author.
>
> For more information on adding line comments to a pull request review, see "Reviewing proposed changes in a pull request."

and

> Note: If you reply to a pull request via email, your comment will be added on the Conversation tab and will not be part of a pull request review.

Which kind of implies that line comments are considered reviews? Very confusing

SebastianMestre:
If you need some reference for what this should support, look into Prolog (the logic programming language) implementations

lushoBarlett:
I'm not particularly lost this time, I just spent like 4 hours of my evening doing integrals and I left this for today.

SebastianMestre:
I mention it because those folk probably have a way better idea of how to deal with this kinda stuff than either one of us

lushoBarlett:
I will skip the type var stuff, because I think that needs to be eliminated, since the meta type system will be able to handle it. But I will leave all further cleanup for later prs. Cleanup so far means stuff like making the TypeFunctionData polymorphic and elimitating the TypeVars.

src/typecheck.cpp
```diff
@@ -40,7 +40,7 @@ void typecheck(TypedAST::ArrayLiteral* ast, TypeChecker& tc) {
 	auto element_type = tc.new_var();
 	for (auto& element : ast->m_elements) {
 		typecheck(element.get(), tc);
-		tc.m_core.unify(element_type, element->m_value_type);
+		tc.m_core.m_mono_core.unify(element_type, element->m_value_type);
```
SebastianMestre, line 43: I don't know if we should put the various unifiers directly in the `TypeChecker`?

SebastianMestre: I literally don't know

src/typesystem.cpp
```diff
@@ -53,55 +53,57 @@ TypeSystemCore::TypeSystemCore() {
 		}
 
 	};
+
+	m_mono_core.unify_function = [&](int a, int b) {
```
SebastianMestre, line 57: Maybe we should change this API so it takes a unifier by reference, instead of having to capture it?

src/typesystem.cpp
```diff
@@ -53,55 +53,57 @@ TypeSystemCore::TypeSystemCore() {
...
+	
+		Unification::Core::TermData& a_data = m_mono_core.term_data[a];
+		Unification::Core::TermData& b_data = m_mono_core.term_data[b];
+
+		m_tf_core.unify(a_data.function_id, b_data.function_id);
+	};
 }
 
 
 void TypeSystemCore::print_type(MonoId mono, int d) {
```
SebastianMestre, line 101: Maybe we should have a print_node in the unifier, instead?

src/typesystem.hpp
```diff
@@ -67,8 +43,18 @@ struct TypeVarData {
...
+	// variable or a term.
+	// We express this variant using an enum, and an index that points
+	// to where the data is in the correct TypeSystemCore vector.
+	//
+	// If type is MonoTag::Var, the data_id index points to a
+	// different mono in TypeSystemCore::mono_data.
+	// It may point to itself, meaning that the type is not known.
+	//
+	// If the type is MonoTag::Term, the index points to a term,
+	// that is stored in TypeSystemCore::term_data
```
SebastianMestre, line 56: outdated comment

src/typechecker.cpp
```diff
@@ -8,9 +8,10 @@
 namespace TypeChecker {
 
 TypeChecker::TypeChecker() {
-	m_meta_core.new_meta(); // 0 | value
-	m_meta_core.new_meta(); // 1 | type func
-	m_meta_core.new_meta(); // 2 | mono type
+	// NOTE: we don't care about the term data
```
SebastianMestre, line 11: more like, "INVARIANT: terms have no arguments"

it is not that we don't care about the arguments, but that we care about the no arguments

src/typesystem.cpp
```diff
@@ -66,6 +66,35 @@ TypeSystemCore::TypeSystemCore() {
...
+			                     : empty;
+
+			std::cerr << "Tried to unify different metatypes.\n"
+			          << "Debug data:\n"
+			          << "i: " << data_i << '\n'
+			          << "j: " << data_j << '\n';
+		}
+#else
+		assert(a == b and "unified two different metatypes");
+#endif
```
SebastianMestre, line 96: we want to assert regardless of the value of DEBUG.

SebastianMestre (changes requested):
There are some things that need addressing, some can be left for later, some are just suggestions.

SebastianMestre:
https://github.com/SebastianMestre/Jasper/blob/67c4c673c6aea0299d81aafcaf1a5324d7547c64/src/typesystem.hpp#L64

Might as well remove this in the same pr, right? like, nothing depends on it. It's a matter of deleting more lines.

src/typecheck.cpp
lushoBarlett: I considered this. I'm not sure either.

Pros:
- more abstraction possibly?
- less accessing through the various components

Cons:
- name clashing (3 billion of each function)

Name clashing is something that triggers me, because I have to rename my functions... It's minor but that's why I left it like that. What do you think?

src/typesystem.cpp
lushoBarlett: I did the first thing for it to work, so I'll take suggestions if you point me to an example to guide me.

src/typesystem.cpp
lushoBarlett: This is a good, yea

src/typechecker.cpp
lushoBarlett: What change do you propose here? I had no good ideas.

src/typesystem.hpp
lushoBarlett: Will delete then.

lushoBarlett:
> Might as well remove this in the same pr

Sure, but I wanted to first make sure that it's actually okay to do it.

src/typechecker.cpp
SebastianMestre: Just a bettsr comment

src/typesystem.cpp
SebastianMestre: Have the std::function take a `unification::core&` as its first argument, then pass `*this` when you call it in `unify`

SebastianMestre: Just an idea (which i do think is better), do whatever you think is better

src/typecheck.cpp
SebastianMestre: what? the member variables are already called `m_mono_something` what would they collide with? what functions are you talking about?

lushoBarlett: Oh I think I understood wrong, you are saying that we put the Unification::Cores in the type checker. Also because there would be so little left in the type system core we could get rid of it altogether?

src/typesystem.cpp
lushoBarlett: It seems unnecessary because I do need to capture `this` anyway, but it's clearer when you are using different cores so I'm leaving it

src/typecheck.cpp
SebastianMestre: yup, that is what I meant. Yeah, we can look at this later.

src/unification.hpp
```diff
@@ -31,7 +31,7 @@ struct Core {
 	void unify(int i, int j);
 
 	int new_var(const char* debug = nullptr);
-	int new_term(int f, std::vector<int> args, const char* debug = nullptr);
+	int new_term(int f = -1, std::vector<int> args = {}, const char* debug = nullptr);
```
SebastianMestre, line 36: remove the default -1 argument. That is not the common usage of this function. We would normally want a compilation error if no arguments are passed.

SebastianMestre (approved):

SebastianMestre:
Good pr. really nice how it manages to abstract 200 lines out of existence.

src/unification.hpp
lushoBarlett: Figured it'd be bad, I wanted to try a method in which we don't have to add the comment explaining why we call with -1 on metatype term creation. I will add it tho

PR #154, opened 2020-09-29
Eval struct expr

SebastianMestre:


SebastianMestre:
as of now, the type checker doesn't complain when checking

    entero := int(<>);

SebastianMestre:
and now, it doesn't complain when checking

    punto := struct {
      x : int(<>);
      y : int(<>);
    };

SebastianMestre:
This also works, by the way, because we always implicitly add `(<>)` to identifiers when we expect a type term

    punto := struct {
      x : int;
      y : int;
    };

SebastianMestre:
This is what the thing I did yesterday was building up to.

src/ct_eval.cpp
```diff
@@ -97,42 +97,103 @@ Own<TypedAST::ReturnStatement> ct_eval(Own<TypedAST::ReturnStatement> ast, TypeC
...
+		for (int i = 0; i < field_count; ++i){
+			MonoId mono = mono_type_from_ast(as_se->m_types[i].get(), tc);
+			std::string name = as_se->m_fields[i].text();
+			assert(!fields.count(name));
+			fields[name] = mono;
+		}
+
+		// TODO: we create a dummy typefunc then make it non-dummy. SERIOUSLY?
+		TypeFunctionId result = tc.m_core.new_dummy_type_function(
+		    TypeFunctionTag::Record, std::move(fields));
```
lushoBarlett, line 126: omg I never realized there was no non-builtin, non-dummy function... I feel very ashamed.

lushoBarlett (approved):
Okay, so I think I get all of this. You traverse the AST and grab any type literals lying around, you make handles with them (not very sure why?) and then you instantiate the types in the core and typecheck them, hence they are now "values in compile time".

E: What I mean with what's in parenthesis is of the likes of, why create handles if you can achieve the same result without them. You don't do anything with them besides contain a little information inside their little scope. Unless I misread or I'm missing something.

SebastianMestre:
Where else do you propose we store their data?

I though about repurposing one of the fields in TypedAST::Declaration, but that's dirtier than I'd like, tbh.

Also, remember that when we eval a type term or a struct expression, one or more of its arguments/field types may not be computed yet.

Example of this last point:

```
pepe := struct {
  a : popo;
};

popo := int;
```

Im not merging yet, maybe there is a better way to do this?

lushoBarlett:
Leaving record:

I missed the part where you assign the handle to the declaration, I thought you only passed the handle to a function, processed it then assign the contents.

I have to better method to do what you did, as of now. So that, to me, indicates that this should be merged now and we will worry about it later.

SebastianMestre:
> Leaving record:
> 
> I missed the part where you assign the handle to the declaration, I thought you only passed the handle to a function, processed it then assign the contents.
> 
> I have to better method to do what you did, as of now. So that, to me, indicates that this should be merged now and we will worry about it later.

got it

PR #158, opened 2020-10-04
Intern strings

SebastianMestre:
Here is the deal: we have strings all over the place. A lot of them are duplicated, which is dumb, and takes up more memory than it should. There is actually a common solution for this: string interning.

What is string interning? It is the act of putting all the strings in your program inside a hash table, and using pointers into hash table entries to refer to particular strings.

Why is this good? Since every string is de-duplicated, we can compare two interned strings by comparing their pointer, meaning all string equality comparisons are O(1). Furthermore, we can use the same pointer as a hash key, meaning string hashing is also O(1) (therefore, things that use unordered_map, like the CTEnv, etc. will be faster) .

In order to implement the string interning operations efficiently, it is necessary to write a custom hash table.

### DIFF
--- a/src/string_set.hpp
+++ b/src/string_set.hpp
@@ -1,6 +1,7 @@
-#include <vector>
+#include <memory>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include <cstdint>
 
@@ -19,7 +20,7 @@ struct StringSet {
 		uint64_t hash_bits : 62;
 	};
 
-	std::vector<std::pair<HashField, std::string>> m_data;
+	std::vector<std::pair<HashField, std::unique_ptr<std::string>>> m_data;
 	size_t m_size;
 
 	StringSet();
@@ -32,8 +33,8 @@ struct StringSet {
 	void insert(char const*);
 
 	bool includes(std::string const&) const;
-	// bool includes(InternedString const&) const;
-	// bool includes(string_view const&) const;
+	// bool includes(InternedString const&) const; // TODO
+	// bool includes(string_view const&) const; // TODO
 	bool includes(char const*, size_t) const;
 	bool includes(char const*) const;
 
--- a/src/string_set.hpp
+++ b/src/string_set.hpp
@@ -3,6 +3,8 @@
 #include <utility>
 #include <vector>
 
+#include "chunked_array.hpp"
+
 #include <cstdint>
 
 struct string_view;
@@ -22,11 +24,14 @@ struct StringSet {
 		static constexpr uint64_t Occupied = 1;
 		static constexpr uint64_t Tombstone = 2;
 
-		uint64_t status : 2 {0};
-		uint64_t hash_bits : 62 {0};
+		std::string* value;
+		uint64_t status : 2;
+		uint64_t hash_bits : 62;
 	};
 
-	std::vector<std::pair<HashField, std::unique_ptr<std::string>>> m_data;
+	
+	ChunkedArray<std::string> m_storage;
+	std::vector<HashField> m_data;
 	size_t m_size {0};
 
 	StringSet();
--- a/src/string_set.hpp
+++ b/src/string_set.hpp
@@ -1,0 +1,45 @@
+#include <vector>
+#include <string>
+#include <utility>
+
+#include <cstdint>
+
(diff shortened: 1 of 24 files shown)


SebastianMestre:
I just realized this actually requires address stability, so the hash table I wrote won't work...

SebastianMestre:
Ok, so, we can actually patch reference stability into it using `unique_ptr`. It's janky, but it's ok.

src/string_set.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+	bool includes(std::string const&) const;
+	// bool includes(InternedString const&) const;
+	// bool includes(string_view const&) const;
+	bool includes(char const*, size_t) const;
+	bool includes(char const*) const;
+
+  private:
+	std::pair<int, bool> insertion_spot(char const*, size_t, uint64_t) const;
+	std::pair<int, bool> find_spot(char const*, size_t, uint64_t) const;
```
lushoBarlett, line 43: These two implementations are identical. Not sure if this is intentional.

src/string_set.hpp
```diff
@@ -0,0 +1,45 @@
...
+	std::vector<std::pair<HashField, std::string>> m_data;
+	size_t m_size;
+
+	StringSet();
+
+	void insert(std::string const&);
+	void insert(std::string&&);
+	// void insert(InternedString const&); // TODO
+	// void insert(string_view const&); // TODO
+	void insert(char const*, size_t);
```
lushoBarlett, line 32: Also identical.

lushoBarlett:
Are you serious GitHub? I commented on the commit storyline instead of the full diff and it still used my comments as reviews. Why is that a thing that happens?? It's so annoying.

src/string_set.hpp
SebastianMestre: They aren't (the loop condition is different), but they could be.

src/string_set.hpp
SebastianMestre: They are not. the `string&&` overload actually moves the string into the hash table

SebastianMestre: The `string const&` can be implemented as a one liner in terms of the `char*, int` overload, though.

SebastianMestre:
I did some changes which actually address the duplication. I am currently fixing some bugs, though.

SebastianMestre:
Lol it works on the CI server

SebastianMestre:
We should probably use interned strings in more places, but I am already pretty tired.

Let me know what you think.

SebastianMestre:
For comparison, here is the output from Valgrind

Before:
```
==17105== Memcheck, a memory error detector
==16686== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==16686== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==16686== Command: ../bin/test_program
==16686== 
==16686== 
==16686== HEAP SUMMARY:
==16686==     in use at exit: 0 bytes in 0 blocks
==16686==   total heap usage: 14,454 allocs, 14,454 frees, 973,984 bytes allocated
==16686== 
```

After:
```
==17105== Memcheck, a memory error detector
==17105== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==17105== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==17105== Command: ../bin/test_program
==17105== 
==17105== 
==17105== HEAP SUMMARY:
==17105==     in use at exit: 0 bytes in 0 blocks
==17105==   total heap usage: 14,475 allocs, 14,475 frees, 900,552 bytes allocated
==17105== 
==17105== All heap blocks were freed -- no leaks are possible
```

That's 73432 fewer bytes.

As of now, I couldn't find any significant differences in execution time, but my guess is that this will pay off in the future. Also, we can handle strings more freely because we know that `InternedString` copies, equality comparisons, and hashes are O(1), which should be a nice quality of life improvement.

lushoBarlett (approved):
Looks fine to me. I skimmed the changes of the sort `std::string -> InternedString`. Note that I found some formatting issues in `string_set.cpp` (lack of spaces before `{` and `(` characters).

PR #159, opened 2020-10-05
implement an ast handler

lushoBarlett:
As we discussed the other day, AST creation is slow. To remedy this, I thought hey, why not have it all in vectors? So that's what this is about.

The implementation (at least a naive approach) was quite simple, although for the _minutiae_ I want some opinions. The central idea is that we have N types of nodes in an AST, so we have N vectors containing each specific type of node. Accessing and creation is done through helper functions using the ASTTag as reference, and I created a helpful indexing struct called ASTHandler::Node that contains the type and the index.

Using this, the AST nodes themselves will not contain `std::unique_ptr<AST*>` but instead `ASTHandler::Node`.

Edit:

As requested, accessing and creating nodes in their respective containers is done with a template interface (which is nicer) and now the AST nodes can store a pointer to the nodes they need.


SebastianMestre:
is it possible to have the API be `Type* create<Type>()` instead of `AST* create(ASTTag)`?

SebastianMestre:
Also, maybe ast allocator would be a better name? I feel like handler means something else

SebastianMestre:
`ASTAllocator<T>` looks good, but `ASTAllocator<T, Ts...>` looks messy.

Why not extract the single type case to a separate class?
```c++
// generic linear allocator for a single type
// all functionality
template<typename T>
struct FooAllocator {
  vector<T> m_data;
  int create();
  T& get(int i);
};

// allocator for a bunch of different types
// no functionality, just a bunch of template glue code
template<typename...Ts>
struct BarAllocator;

template<>
struct BarAllocator<> {
  template<typename U>  
  int create(){ static_assert(false, "No matching type"); }
};

template<typename T, typename...Ts>
struct BarAllocator<T, Ts...> {
  FooAllocator<T> m_data;
  BarAllocator<Ts...> m_rest;

  template<typename U>
  enable_if<is_same_v<T,U>, int>
  create() { return m_data.create(); }

  template<typename U>
  enable_if<!is_same_v<T,U>, int>
  create() { return m_rest.template create<U>(); }  
};
```

I'm not sure I got the code right, but you get the idea

lushoBarlett:
Looks like a decent pattern to use when doing template recursion in general actually. I'll take your advice here and keep it in mind in the future.

SebastianMestre:
I don't know how you intend this to be used, but if the common usage is storing pointers, this is currently broken due to the lack of address stability for values in a std::vector (that is, if you have a pointer into the buffer of a vector, and the vector reallocates, now that pointer is no longer valid)

lushoBarlett:
This pr will also try to replace every heap allocation of a node with a `NodeAllocator::make<AST>();`. Both for `AST`s and `TypedAST`s. So prepare for a huge diff from this point onwards.

SebastianMestre:
It'd be neat to get the 'no matching type' errors at compile time; here is a godbolt link with a possible way to do just that https://godbolt.org/z/T81zsr

lushoBarlett:
Sorry for the large diff, the "must compile on commit" rule is a cruel mistress

src/ast.hpp
```diff
@@ -145,90 +145,90 @@ struct Identifier : public AST {
...
 	    : AST {ASTTag::ReturnStatement} {}
 };
 
 struct IfElseStatement : public AST {
-	Own<AST> m_condition;
-	Own<AST> m_body;
-	Own<AST> m_else_body; // can be nullptr
+	AST* m_condition;
+	AST* m_body;
+	AST* m_else_body; // can be nullptr
```
SebastianMestre, line 205: Pointers don't get automatically initialized to nullptr, so a default value should be added.

SebastianMestre: This is not the only place where this applies.

src/node_allocator.hpp
```diff
@@ -42,4 +44,9 @@ struct NodeAllocator<T, Ts...> {
 	template <typename U>
```
SebastianMestre, line 44: Should probably delete this class' copy constructor (you will then need to default the default and move constructors)

src/parse.cpp
```diff
@@ -4,30 +4,36 @@
...
 
+	std::pair<AST::AST*, AST::Allocator> graph {nullptr, {}};
+
 	Parser p;
 	p.m_lexer = &l;
+	p.m_ast_allocator = &graph.second;
 
-	return p.parse_top_level();
+	auto result = p.parse_top_level();
+	return {result.m_error, graph};
```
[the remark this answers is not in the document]
SebastianMestre: Also, you're missing

    graph.first = result.m_result;

SebastianMestre (changes requested):

src/ast.hpp
lushoBarlett: Funny enough, the bug I just fixed was exactly that

src/node_allocator.hpp
lushoBarlett: I did it for debugging and I decided to leave it in already

src/parse.cpp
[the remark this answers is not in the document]
lushoBarlett: Took your advice, and this is no longer the approach used

src/ast_tag.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "node_allocator.hpp"
```
SebastianMestre, line 3: why would `ast_tag.hpp` need `node_allocator.hpp"?

src/desugar.cpp
```diff
@@ -2,184 +2,185 @@
...
+		ast->m_value = desugar(ast->m_value, alloc);
 
 	return ast;
 }
 
-Own<AST> desugar(Own<DeclarationList> ast) {
-	for (auto&& declaration : ast->m_declarations)
-		declaration = desugar(std::move(declaration));
+AST* desugar(DeclarationList* ast, Allocator& alloc) {
+	for (auto declaration : ast->m_declarations)
```
SebastianMestre, line 21: this should iterate by reference

src/desugar.cpp
```diff
@@ -2,184 +2,185 @@
...
+		ast->m_body[i] = desugar(ast->m_body[i], alloc);
 
 	return ast;
 }
 
-Own<AST> desugar(Own<FunctionLiteral> ast) {
-	for (auto&& arg : ast->m_args)
-		arg = desugar(std::move(arg));
+AST* desugar(FunctionLiteral* ast, Allocator& alloc) {
+	for (auto arg : ast->m_args)
```
SebastianMestre, line 53: this should iterate by reference

SebastianMestre (changes requested):

src/ast_tag.hpp
lushoBarlett: That was an oopsie

src/desugar.cpp
lushoBarlett: Also fixed it just now

SebastianMestre:
looking good! would you try your hand at some benchmarks?

There are two ways you can do this: micro benchmarks (e.g. a tight loop that calls `make_unique` for a bunch of different ast types and another one that uses `NodeAllocator.make`) and macro benchmarks (e.g. feed a large input into the parser before and after the change, and measure how long it takes overall)

Micro benchmarks are easier to do, but it's hard to justify a big change like this without evidence that it has an impact on the overall process, so, a macro benchmark would be nicer.

SebastianMestre:
I pulled, measured, and found a 17.92% mean runtime improvement with a p-value of 0.0054. Very cool.

src/node_allocator.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "chunked_array.hpp"
+
+#include <iostream>
+
+#include <type_traits>
+
+template <typename T>
+struct SingleAllocator {
```
SebastianMestre, line 10: yeh my man that's pretty fast

SebastianMestre (approved):

lushoBarlett:
All right, I'll merge this and do the rest in another pr

PR #161, opened 2020-10-08
Fast hash table

SebastianMestre:
In my benchmarks, the slowest thing in the runtime was, by far, the hash table. Here, I strive to make it faster.


lushoBarlett (approved):
Noticed the slowness of `unordered_map` too when I did the benchmark, I'm familiar with this code when I did the review for `InternedString` so it's okay.

SebastianMestre:
this was kinda fun. I observed a (roughly) 12% improvement in a simple test program (source below).

This program implements a simple binary counter. The important part is that is does `2**15` iterations of the outer loop, meaning we get a nice and large amount of samples on `perf`

```rust
func := fn(a,b,c,d,e,f,g,h,i,j,k,l,m) => "b";
__invoke := fn() {
	x0 := 0;
	x1 := 0;
	x2 := 0;
	// and so on until xf
	iters := 0;
	s1 := "a";
	while (xf == 0){
		iters = iters + 1;
		s1 = func(s1,s1,s1,s1,s1,s1,s1,s1,s1,s1,s1,s1,s1);
		if(x0 == 1){
			if(x1 == 1){
				if(x2 == 1){
					// and so on until xe
					x3 = 1 - x3;
				}
				x2 = 1 - x2;
			}
			x1 = 1 - x1;
		}
		x0 = 1 - x0;
	}
	return iters;
};
```

There are more optimizations that can be done, but this is already pretty nice.

SebastianMestre:
should I remove `FlatMap`? We don't use it currently, but I think it's a pretty nice hash table, and could come in handy in the future.

SebastianMestre:
Actually, forget it. The hash tables I wrote are dumb, and actually slower than `std::set` for the small amount of elements that typically are in a single scope

PR #162, opened 2020-10-08
Typed ast allocator

lushoBarlett:
This is the sequel of the last one I did, where we replace `unique_ptr`s for a `NodeAllocator<...>`.

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -11,89 +11,98 @@ namespace TypeChecker {
 
 // literals
 
-Own<TypedAST::ArrayLiteral> ct_eval(Own<TypedAST::ArrayLiteral> ast, TypeChecker& tc) {
-	for (auto&& element : ast->m_elements)
-		element = ct_eval(std::move(element), tc);
+TypedAST::ArrayLiteral* ct_eval(
+    TypedAST::ArrayLiteral* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	for (auto& element : ast->m_elements)
+		element = ct_eval(element, tc, alloc);
 	return ast;
 }
 
 // expressions
 
-Own<TypedAST::FunctionLiteral> ct_eval(
-    Own<TypedAST::FunctionLiteral> ast, TypeChecker& tc) {
+TypedAST::FunctionLiteral* ct_eval(
+    TypedAST::FunctionLiteral* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
 
 	// TODO: type hints
 
 	assert(ast->m_body->type() == TypedASTTag::Block);
-	auto body = static_cast<TypedAST::Block*>(ast->m_body.get());
-	for (auto&& child : body->m_body)
-		child = ct_eval(std::move(child), tc);
+	auto body = static_cast<TypedAST::Block*>(ast->m_body);
+	for (auto child : body->m_body)
+		child = ct_eval(child, tc, alloc);
 
 	return ast;
 }
 
-Own<TypedAST::CallExpression> ct_eval(
-    Own<TypedAST::CallExpression> ast, TypeChecker& tc) {
+TypedAST::CallExpression* ct_eval(
+    TypedAST::CallExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
 
-	ast->m_callee = ct_eval(std::move(ast->m_callee), tc);
+	ast->m_callee = ct_eval(ast->m_callee, tc, alloc);
 
-	for (auto&& arg : ast->m_args)
-		arg = ct_eval(std::move(arg), tc);
-
-	return ast;
-}
-
-Own<TypedAST::IndexExpression> ct_eval(Own<TypedAST::IndexExpression> ast, TypeChecker& tc) {
-	ast->m_callee = ct_eval(std::move(ast->m_callee), tc);
-	ast->m_index = ct_eval(std::move(ast->m_index), tc);
+	for (auto& arg : ast->m_args)
+		arg = ct_eval(arg, tc, alloc);
 
 	return ast;
 }
 
-Own<TypedAST::TernaryExpression> ct_eval(Own<TypedAST::TernaryExpression> ast, TypeChecker& tc) {
-	ast->m_condition = ct_eval(std::move(ast->m_condition), tc);
-	ast->m_then_expr = ct_eval(std::move(ast->m_then_expr), tc);
-	ast->m_else_expr = ct_eval(std::move(ast->m_else_expr), tc);
+TypedAST::IndexExpression* ct_eval(
+    TypedAST::IndexExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	ast->m_callee = ct_eval(ast->m_callee, tc, alloc);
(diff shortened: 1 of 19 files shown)


src/ct_eval.cpp
```diff
@@ -11,89 +11,98 @@ namespace TypeChecker {
...
+    TypedAST::FunctionLiteral* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
 
 	// TODO: type hints
 
 	assert(ast->m_body->type() == TypedASTTag::Block);
-	auto body = static_cast<TypedAST::Block*>(ast->m_body.get());
-	for (auto&& child : body->m_body)
-		child = ct_eval(std::move(child), tc);
+	auto body = static_cast<TypedAST::Block*>(ast->m_body);
+	for (auto child : body->m_body)
```
SebastianMestre, line 30: iterate by ref

src/typed_ast.hpp
```diff
@@ -166,90 +168,90 @@ struct Identifier : public TypedAST {
...
+	TypedAST* m_body;
 
 	WhileStatement()
 	    : TypedAST {TypedASTTag::WhileStatement} {}
 };
 
 struct StructExpression : public TypedAST {
-	std::vector<Identifier> m_fields;
-	std::vector<Own<TypedAST>> m_types;
+	std::vector<Identifier*> m_fields;
```
SebastianMestre, line 245: why are these pointers now?

src/typed_ast.hpp
```diff
@@ -60,7 +62,7 @@ struct Declaration : public TypedAST {
 
 // doesnt have a ast_vtype
 struct DeclarationList : public TypedAST {
-	std::vector<Own<Declaration>> m_declarations;
+	std::vector<Declaration*> m_declarations;
```
SebastianMestre, line 65: I feel like this indirection is not necessary, but I'm not gonna bother you with that now.

src/typed_ast.hpp
```diff
@@ -128,23 +130,23 @@ struct ObjectLiteral : public TypedAST {
...
 	DictionaryLiteral()
 	    : TypedAST {TypedASTTag::DictionaryLiteral} {}
 };
 
 struct FunctionLiteral : public TypedAST {
 	MonoId m_return_type;
-	Own<TypedAST> m_body;
-	std::vector<Declaration> m_args;
+	TypedAST* m_body;
+	std::vector<Declaration*> m_args;
```
SebastianMestre, line 149: Why are these pointers now?

src/typed_ast.cpp
```diff
@@ -4,222 +4,225 @@
...
 	for (auto arg : ast->m_args) {
 		assert(arg->type() == ASTTag::Declaration);
-		auto* decl = static_cast<AST::Declaration*>(arg);
+		auto decl = static_cast<AST::Declaration*>(arg);
 
-		Declaration typed_decl;
-		typed_decl.m_identifier_token = decl->m_identifier_token;
-		typed_decl.m_surrounding_function = typed_function.get();
-		typed_function->m_args.push_back(std::move(typed_decl));
+		auto typed_decl = alloc.make<Declaration>();
```
SebastianMestre, line 66: Why did you change this one from a value to a pointer?

SebastianMestre (approved):
I only skimmed it (I'm pretty tired, sorry). It looks mostly ok, but I found some places where you changed by-reference iteration to by-value (probably bugs), and some places where you changed some by-value members to by-pointer (probably slower). I don't understand why, and would like to know your reasoning.

src/typed_ast.hpp
lushoBarlett: I figured since they are now allocated elsewhere I can't store them here directly, another approach would be to exclude some of the nodes in the node allocator and have them stored here wherever they are needed.

src/typed_ast.cpp
lushoBarlett: This one, honestly, it was a matter of consistency to have them all in the allocator... I guess there are many places where that could be avoided just by not allocating them in the node allocator. I should rethink that, I also did it in the last pr.

lushoBarlett:
I'll dedicate the last of this pr to reverting some things that you pointed out, mainly avoid storing in the allocator when not necessary. I also noticed some places where we could replace `::AST*` with `::Declaration` (in functions), so let me know if you want me to also do that.

src/ct_eval.cpp
lushoBarlett: This one I completely missed it. Probably a misclick in vim.

src/typed_ast.cpp
SebastianMestre: As a rule of thumb, the less indirection the better :)

SebastianMestre:
> I'll dedicate the last of this pr to reverting some things that you pointed out, mainly avoid storing in the allocator when not necessary. I also noticed some places where we could replace `::AST*` with `::Declaration` (in functions), so let me know if you want me to also do that.

That'd be nice

src/ct_eval.cpp
```diff
@@ -86,7 +86,7 @@ TypedAST::IfElseStatement* ct_eval(
 
 TypedAST::ForStatement* ct_eval(
     TypedAST::ForStatement* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
-	ast->m_declaration = ct_eval(ast->m_declaration, tc, alloc);
+	ct_eval(&ast->m_declaration, tc, alloc);
```
SebastianMestre, line 89: we need to keep this in mind, in case `ct_eval` starts doing more complicated things.

src/desugar.cpp
```diff
@@ -66,8 +66,10 @@ AST* desugar(ShortFunctionLiteral* ast, Allocator& alloc) {
 	block->m_body.push_back(return_stmt);
 
 	auto func = alloc.make<FunctionLiteral>();
-	for (auto& arg : ast->m_args)
-		func->m_args.push_back(desugar(arg, alloc));
+	func->m_args = std::move(ast->m_args);
+
+	for (auto& arg : func->m_args)
+		desugar(&arg, alloc);
```
SebastianMestre, line 72: ah yeah, the way it was before was pretty dumb

SebastianMestre: this is way nicer

src/typed_ast.cpp
```diff
@@ -59,24 +60,21 @@ TypedAST* convert_ast(AST::DictionaryLiteral* ast, Allocator& alloc) {
...
+		typed_function->m_args.push_back(std::move(typed_decl));
 	}
 
 	typed_function->m_body = convert_ast(ast->m_body, alloc);
 
 	return typed_function;
 }
 
-Declaration* convert_ast(AST::Declaration* ast, Allocator& alloc) {
+TypedAST* convert_ast(AST::Declaration* ast, Allocator& alloc) {
```
SebastianMestre: That's weird.. why does this not return a `Declaration*`?

SebastianMestre (approved):
Looks good

src/typed_ast.cpp
lushoBarlett: I reckon there was a type error. I supposed that it was because I was changing the return type. I find it weird too because I was sure it was like that before.

PR #163, opened 2020-10-09
String set improvements

SebastianMestre:
I had written this hash table, and I was pretty happy with it, but, looking at it a second time... man... it was not very nice.

It had a pretty big bug: the hashes that `compute_effective_hash` returned weren't actually 62 bit hashes (well, they were, but the 0 bits were not the highest order ones, but ones in the middle).

It also didn't use the 62 bits of hash that we store to save on string comparisons, which is a pretty massive oversight.

Other than that, there was (and still is, to a lesser degree) general clumsiness to how the code is written.

Another thing that could be improved is that `StringSet::includes` can be written more efficiently by not looking for free slots during the scan. I didn't do this because we don't use it anywhere outside of tests, so it's not really worth the extra code.


lushoBarlett (approved):
Looks much nicer this way!

PR #165, opened 2020-10-11
implement stack-based local variables

SebastianMestre:
I figured that, instead of spending more work on optimizing the dictionary in Interpreter::Environment, we should do something to forego it entirely, just like in compiled languages.

This is a first pass on that concept.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -7,6 +7,7 @@ OBJDIR := ../build
 OBJNAMES := \
 	ast.o \
 	compile_time_environment.o \
+	compute_offsets.o \
 	ct_eval.o \
 	desugar.o \
 	error_report.o \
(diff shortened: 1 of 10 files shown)


SebastianMestre:
For now, we still need to build the scope stack because this doesnt handle captures. Handling captures is what comes next.

Once we handle captures, we no longer need a scope stack, just a single global scope.

After that, we can expand this to work with globals, too. At that point, we could get rid of Interpreter::Scope entirely.

SebastianMestre:
It's essentially working now

SebastianMestre:
It still has a very 'hacked-on' feeling to it, so it needs more clean up, but I think this is pretty good, I'll run some benchmarks tomorrow.

SebastianMestre:
I just measured, I observed an 18% decrease in mean execution time (standard deviation is about 0.1%)

SebastianMestre:
This actually opens up some changes that could be done to simplify and optimize the runtime even more, but this is already a pretty large patch.

lushoBarlett (approved):
Looks fine to me, honestly

PR #166, opened 2020-10-13
Optimize argument passing

SebastianMestre:
I figured we don't really need to put arguments in a vector before we pass them. So, I am trying to make it so.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -17,6 +17,7 @@ OBJNAMES := \
 	metacheck.o \
 	parser.o \
 	parse.o \
+	span.o \
 	string_set.o \
 	string_view.o \
 	tarjan_solver.o \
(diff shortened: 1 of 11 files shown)


SebastianMestre:
I haven't managed to get rid of the vector yet, but I did simplify some things, and it did get faster.

SebastianMestre:
Second round of measurements:



before (master) (30 runs):

    0,114305 +- 0,000115 seconds time elapsed ( +- 0,10% )

before (round 1 on this pr) (30 runs):

    0,104121 +- 0,000782 seconds time elapsed ( +- 0,75% )

now (30 runs):

    0,091157 +- 0,000209 seconds time elapsed  ( +-  0,23% )

overall improvement: 20.25%

SebastianMestre:
Funnily enough, If you take out `Span`, the change to `gc_ptr`, and the extra line in the Makefile, this is actually 1 line shorter

lushoBarlett (approved):
This is a clever approach, I think what it does is:

The call expression pushes the arguments, then the function itself already points to those locations. That saves a whole vector every function.

Cool stuff!

PR #167, opened 2020-10-13
Actually run the gc during execution

SebastianMestre:
We were not really running the gc before. So, now we are.

I was incredibly surprised to see that this is actually faster than before for programs that allocate a lot. I don't exactly know how to explain it. I will say that this gave me a 19.91% (!!!) speedup (relative to master) on the benchmark I used on the other PRs, though.

### DIFF
--- a/src/interpreter/environment.cpp
+++ b/src/interpreter/environment.cpp
@@ -34,19 +34,6 @@ Value* Environment::fetch_return_value() {
 	return rv;
 }
 
-void Environment::run_gc() {
-	m_gc->unmark_all();
-	m_gc->mark_roots();
-
-	for (auto p : m_stack)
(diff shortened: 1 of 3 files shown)


lushoBarlett (approved):
Wow, that's a little odd honestly. I guess it's because memory being freed means malloc takes less. Perhaps Linux does something there to speed that up.

SebastianMestre:
After merging the other improvements from master, I got these timings (30 runs):

    0,066685 +- 0,000226 seconds time elapsed  ( +-  0,34% )

Remember that, before these two PRs, the timings were (30 runs):

    0,114305 +- 0,000115 seconds time elapsed ( +- 0,10% )

This implies a whopping 41.66% improvement!

PR #168, opened 2020-10-14
Have the GC manage its own memory

SebastianMestre:
Right now, we put every object on its own system-level allocation (i.e. we use `new T` to allocate it). Instead, we could have our own slab block allocators that put many objects on the same buffer. This should be a lot more efficient.

I started working towards that, but it's still in very early stages.


SebastianMestre:
This is probably way too stale to be of any use...

PR #169, opened 2020-10-15
Compute number and integer literal values during ast conversion

SebastianMestre:
It's not like string to number conversions are super slow or anything, but it's plain wasteful not to do the conversions statically.

Of course, I measured.

Timings before:

    0.067384 +- 0.000311 seconds

Timings after:

    0.063496 +- 0.000713 seconds

It is a 5.77% improvement on the benchmark I've been using.


lushoBarlett (approved):
It's better this way for value manipulation in C++

PR #170, opened 2020-10-16
Use the stack more

SebastianMestre:
So, I figured things would be simpler if we put the result of evaluating expressions on the stack, instead of returning them from eval.

I still think this is the case, we just can't see it yet because it needs some refactoring. Right now, we do use the stack, but then immediately pop values and put them in `gc_ptr`s, which is what we were doing before, but worse.

I'll keep working on this for a bit, I'm sure it only got worse before it gets better.

### DIFF
--- a/src/interpreter/environment.cpp
+++ b/src/interpreter/environment.cpp
@@ -132,10 +132,9 @@ void Environment::push_boolean(bool b) {
 	run_gc_if_needed();
 }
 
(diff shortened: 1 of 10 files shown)


SebastianMestre:
I'll be busy later, so I'm setting this as ready for review

lushoBarlett (approved):
All right that wasn't so bad, it was all the same change essentially

PR #171, opened 2020-10-17
Rewrite the Makefile

SebastianMestre:
I rewrote the Makefile. I think it is not only nicer to read, but also nicer to use.

You can now specify a build mode. The options are 'debug' (for debugging) and 'tuning' (for performance work). If no mode is provided, or the mode is not recognized, release mode is assumed.

usage examples

```shell
make interpreter MODE=tuning
make tests MODE=debug
make -j # implicitly a release (optimized) build
```

The various build modes set different compiler flags, suitable for their purposes. They put their build files in separate directories, as to avoid having to rebuild from scratch when going from one build mode to another.

Also: the target `test_program` is now called `tests`, the executable `run_tests`, and the build pretty-prints each file it starts to build.


lushoBarlett (approved):
That's new tech! Amazing. Also it's great that you moved the Makefile to root

PR #174, opened 2020-10-20
Fix a copypaste error in the build

SebastianMestre:
When I overhauled the build, I took the sed command from the old build, and it ended up with a stray $(OBJDIR) being left in.

Replacing it with $(BUILD_DIR) should do the trick.


lushoBarlett (approved):
Good thing you found the bug

PR #175, opened 2020-10-20
Separate things into directories

SebastianMestre:
I figured it might be nice to put stuff in directories. I don't really know what would be a good  way to organize things, but I started by putting all the helper data structures into a `src/utils` subdirectory


lushoBarlett:
Maybe Tarjan Solver could also be moved to utils? At least it's for sure a separate thing.

Also, is it super wrong to put all the ast stuff in an ast folder, and the typed ast stuff in a typed ast folder? I feel like root folder could go around with having only the "phases". But it's just a thought.

SebastianMestre:
> Maybe Tarjan Solver could also be moved to utils? At least it's for sure a separate thing.
yeah

> Also, is it super wrong to put all the ast stuff in an ast folder, and the typed ast stuff in a typed ast folder? I feel like root folder could go around with having only the "phases". But it's just a thought.
uhhhh... I don't know. I don't like that.

lushoBarlett (approved):
It's an improvement over how it was before for sure

Issue #155, opened 2020-09-30
There is no regular type function construction method


lushoBarlett:
In the typesystem, there's a constructor for dummy type functions. We need a method just like that, for regular function.

Perhaps a good way would be to add a parameter at the end to make the function a dummy or not?

PR #176, opened 2020-10-20
Value constructors parse and typecheck

lushoBarlett:
Implements construction of values from types, like `pair { 1; 2; }` or maybe `int { 0; }`.

Also implements addition of non-dummy type functions, mentioned in issue #155.

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -170,12 +170,12 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 	} else if (ast->type() == TypedASTTag::StructExpression) {
 		auto as_se = static_cast<TypedAST::StructExpression*>(ast);
 
-		std::vector<std::string> fields;
-		std::unordered_map<std::string, MonoId> structure;
+		std::vector<InternedString> fields;
+		std::unordered_map<InternedString, MonoId> structure;
 		int field_count = as_se->m_fields.size();
 		for (int i = 0; i < field_count; ++i){
 			MonoId mono = mono_type_from_ast(as_se->m_types[i], tc);
-			std::string name = as_se->m_fields[i].text().str();
+			InternedString name(as_se->m_fields[i].text().str());
 			assert(!structure.count(name));
 			structure[name] = mono;
 			fields.push_back(name);
@@ -265,6 +265,14 @@ TypedAST::DeclarationList* ct_eval(
 	return ast;
 }
 
+TypedAST::MonoTypeHandle* ct_eval(
+    TypedAST::TypeTerm* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	auto handle = alloc.make<TypedAST::MonoTypeHandle>();
+	handle->m_value = mono_type_from_ast(ast, tc);
+	handle->m_syntax = ast;
+	return handle;
+}
+
 TypedAST::TypedAST* ct_eval(
     TypedAST::TypedAST* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
 #define DISPATCH(type)                                                         \
@@ -299,6 +307,8 @@ TypedAST::TypedAST* ct_eval(
 
 		DISPATCH(Declaration);
 		DISPATCH(DeclarationList);
+
+		DISPATCH(TypeTerm);
 	}
 
 	std::cerr << "Unhandled case in " << __PRETTY_FUNCTION__ << " : "
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -175,7 +175,7 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 		int field_count = as_se->m_fields.size();
 		for (int i = 0; i < field_count; ++i){
 			MonoId mono = mono_type_from_ast(as_se->m_types[i], tc);
-			InternedString name(as_se->m_fields[i].text().str());
+			InternedString name = as_se->m_fields[i].text();
 			assert(!structure.count(name));
 			structure[name] = mono;
 			fields.push_back(name);
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -66,6 +66,16 @@ TypedAST::RecordAccessExpression* ct_eval(
 	return ast;
 }
 
+TypedAST::ConstructorExpression* ct_eval(
+    TypedAST::ConstructorExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	ast->m_constructor = ct_eval(ast->m_constructor, tc, alloc);
+
+	for (auto& arg : ast->m_args)
+		arg = ct_eval(arg, tc, alloc);
+
+	return ast;
+}
+
 // statements
 
 TypedAST::Block* ct_eval(
@@ -238,6 +248,7 @@ TypedAST::TypedAST* ct_eval(
 		DISPATCH(IndexExpression);
 		DISPATCH(TernaryExpression);
 		DISPATCH(RecordAccessExpression);
+		DISPATCH(ConstructorExpression);
 
 		DISPATCH(Block);
 		DISPATCH(IfElseStatement);
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -142,11 +142,8 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 			fields.push_back(name);
 		}
 
-		// TODO: we create a dummy typefunc then make it non-dummy. SERIOUSLY?
-		TypeFunctionId result = tc.m_core.new_dummy_type_function(
+		TypeFunctionId result = tc.m_core.new_type_function(
 		    TypeFunctionTag::Record, std::move(fields), std::move(structure));
-
-		tc.m_core.m_type_functions[result].is_dummy = false;
 
 		return result;
 	} else {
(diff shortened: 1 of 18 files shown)


lushoBarlett:
I decided to end the pr here for a couple reasons:
- eval phase is bad right now, like _really_ bad for this feature to exist.
- even if I manage to somehow do it, I would have done so many other things... and scope creep y'know?

SebastianMestre (changes requested):
Overall, I see what you're trying to do. It mostly makes sense to me.

What does confuse me is how you store a TypeFunctionId in Identifier::m_value_type.

I feel like this is either an ugly hack which could be a source of many bugs, or a mistake spawned from a lack of understanding.

If it's the latter, I can talk you through the way the concepts fit together. If it's the former, please, find an alternative solution, or leave a very nice comment explaining the reasoning behind this.

src/typecheck.cpp
```diff
@@ -236,13 +237,39 @@ void typecheck(TypedAST::RecordAccessExpression* ast, TypeChecker& tc) {
...
+
+		MonoId field_type = tf_data.structure[tf_data.fields[i]];
+		tc.m_core.m_mono_core.unify(field_type, ast->m_args[i]->m_value_type);
+	}
+
+	// NOTE: this is a special case for the `name { }` syntax sugar,
+	// but perhaps the left side should only be a TypeTerm and not possibly an Identifier?
+	if (ast->m_constructor->m_meta_type == tc.meta_typefunc())
+		ast->m_value_type =
+		    tc.m_core.new_term(ast->m_constructor->m_value_type, {}, "record instance");
```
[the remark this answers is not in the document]
SebastianMestre: Also, once we reach `typecheck`, there should be no `TypedAST::TypeTerm`s anymore, they should all have been turned into `MonoTypeHandle`s

SebastianMestre:
I didn't think of this while reviewing, but:

Since ct_eval converts StructExpressions, and TypeTerms into their TC representation, we should never hit either one of them in `typecheck`. This fact will probably clear up some doubts you may have.


A better design would be to have ct_eval(Identifier) return a MonoTypeHandle, or a TypeFunctionHandle, whenever it finds that `m_meta_type` is not `meta_value`. This should make dealing with stuff in later stages a bit nicer. (because we wouldn't need to special case identifiers)

src/ct_eval.cpp
```diff
@@ -121,20 +131,19 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 	} else if (ast->type() == TypedASTTag::StructExpression) {
 		auto as_se = static_cast<TypedAST::StructExpression*>(ast);
 
-		std::unordered_map<std::string, MonoId> fields;
+		std::vector<std::string> fields;
+		std::unordered_map<std::string, MonoId> structure;
```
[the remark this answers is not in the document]
lushoBarlett: Oh yeah absolutely. Fields do not change.

src/typecheck.cpp
```diff
@@ -236,13 +237,39 @@ void typecheck(TypedAST::RecordAccessExpression* ast, TypeChecker& tc) {
 	MonoId member_type = tc.new_var();
 	ast->m_value_type = member_type;
 
-	TypeFunctionId dummy_tf = tc.m_core.new_dummy_type_function(
-	    TypeFunctionTag::Record, {{ast->m_member->m_text.str(), member_type}});
+	TypeFunctionId dummy_tf = tc.m_core.new_type_function
+	    ( TypeFunctionTag::Record
+	    , {ast->m_member->m_text.str()}
```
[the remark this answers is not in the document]
lushoBarlett: Yea I should revisit the implementation of type function unification actually

src/typecheck.cpp
```diff
@@ -95,9 +95,10 @@ void typecheck(TypedAST::Identifier* ast, TypeChecker& tc) {
 		                        ? tc.m_core.inst_fresh(declaration->m_decl_type)
 		                        : declaration->m_value_type;
 	} else if (meta_type == tc.meta_typefunc()) {
-		assert(0 && "Accessed a name of a typefunc (not implemented)");
-		// we are a type function.
-		// TODO: not too sure what needs to be done...
+		// simply match with its declaration
+		// TODO: do 'inst fresh' but for type functions
+		auto handle = static_cast<TypedAST::TypeFunctionHandle*>(declaration->m_value);
+		ast->m_value_type = handle->m_value;
```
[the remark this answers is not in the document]
lushoBarlett: This mistake was due to confusion

src/typecheck.cpp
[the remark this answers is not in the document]
SebastianMestre: I was wrong on this comment, BTW.

This part in particular:

> so is assuming that if the metatype is typefunc, then the node is an identifier (e.g. it could be a StructExpression).

It could only be either a TypeFunctionHandle, or an Identifier (Only a Handle with my latest patch)

src/ct_eval.cpp
```diff
@@ -170,12 +170,12 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
...
 
-		std::vector<std::string> fields;
-		std::unordered_map<std::string, MonoId> structure;
+		std::vector<InternedString> fields;
+		std::unordered_map<InternedString, MonoId> structure;
 		int field_count = as_se->m_fields.size();
 		for (int i = 0; i < field_count; ++i){
 			MonoId mono = mono_type_from_ast(as_se->m_types[i], tc);
-			std::string name = as_se->m_fields[i].text().str();
+			InternedString name(as_se->m_fields[i].text().str());
```
SebastianMestre, line 178: `Identifier::text` returns an InternedString. you don't need to call `.str()` here

```c++
auto name = as_se->m_fields[i].text();
```

should do the same thing (if a little bit faster)

src/ct_eval.cpp
```diff
@@ -265,6 +265,14 @@ TypedAST::DeclarationList* ct_eval(
 	return ast;
 }
 
+TypedAST::MonoTypeHandle* ct_eval(
+    TypedAST::TypeTerm* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	auto handle = alloc.make<TypedAST::MonoTypeHandle>();
+	handle->m_value = mono_type_from_ast(ast, tc);
+	handle->m_syntax = ast;
+	return handle;
+}
```
SebastianMestre, line 274: Oh I though I had already written this. Silly me

src/typecheck.cpp
```diff
@@ -230,8 +230,8 @@ void typecheck(TypedAST::RecordAccessExpression* ast, TypeChecker& tc) {
 
 	TypeFunctionId dummy_tf = tc.m_core.new_type_function
 	    ( TypeFunctionTag::Record
-	    , {ast->m_member->m_text.str()}
-	    , {{ast->m_member->m_text.str(), member_type}}
+	    , {} // we don't care for fields in dummies
+	    , {{InternedString(ast->m_member->m_text.str()), member_type}}
```
SebastianMestre, line 234: `Token::m_text` is actually an InternedString, there is no need to call `.str()` and then make a new `InternedString`. `ast->m_member->m_text` suffices

NIT: "we don't care *about* fields in dummies"

"not care for something" means to dislike that thing

src/typecheck.cpp
```diff
@@ -240,10 +240,12 @@ void typecheck(TypedAST::RecordAccessExpression* ast, TypeChecker& tc) {
 
 void typecheck(TypedAST::ConstructorExpression* ast, TypeChecker& tc) {
 	typecheck(ast->m_constructor, tc);
-	TypeFunctionId tf = tc.m_core.m_tf_core.find_term(ast->m_constructor->m_value_type);
-	TypeFunctionData& tf_data = tc.m_core.m_type_functions[tf];
 
-	// TODO: match type arguments as well, we only support 0 now
+	auto handle = static_cast<TypedAST::MonoTypeHandle*>(ast->m_constructor);
+	assert(handle && "Error: Value constructor was not handled");
```
SebastianMestre, line 245: static cast doesn't return null if you perform an runtime-invalid cast. it wrongly succeedes without warning.

did you mean to `assert(ast->m_constructor->type() == TypedASTTag::MonoTypeHandle)` ?

src/typecheck.cpp
```diff
@@ -240,10 +240,12 @@ void typecheck(TypedAST::RecordAccessExpression* ast, TypeChecker& tc) {
...
+
+	TypeFunctionId tf = tc.m_core.m_mono_core.find_function(handle->m_value);
+	TypeFunctionData& tf_data = tc.m_core.m_type_functions[tf];
```
SebastianMestre, line 248: Is this right? I can't seem to keep track of how typefuncs are stored after the unification refactor

SebastianMestre (approved):
Looks good. I spotted a few places where InternedString could be leveraged a bit more. Up to you if you want to address those.

src/typecheck.cpp
SebastianMestre: NIT: "we don't care *about* fields in dummies"

"not care for something" means to dislike that thing

src/typecheck.cpp
lushoBarlett: A monotype handle stores a MonoId as its value. At the `m_mono_core`, the terms are applications of type functions with some mono arguments, so I use the very handy `find_function` to get the function directly (and not the term). That is a TypeFunctionId, so I fetch its data (that has been separated somewhere else for not being term-like) in the special `m_type_functions` vector.

SebastianMestre: And TypeFunctionId is an index into that array? And not into `m_core.m_tf_core.terms`?

PR #177, opened 2020-10-22
have ct_eval(Identifier) return a handle if appropiate

SebastianMestre:
@lushoBarlett 

This patch makes what you are working on a bit simpler, I think. Hopefully you can make use of it!

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -136,7 +136,9 @@ TypedAST::ReturnStatement* ct_eval(
 
 TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 	assert(tc.m_core.m_meta_core.find(ast->m_meta_type) == tc.meta_typefunc());
-	if(ast->type() == TypedASTTag::Identifier){
+	if (ast->type() == TypedASTTag::TypeFunctionHandle) {
+		return static_cast<TypedAST::TypeFunctionHandle*>(ast)->m_value;
+	} else if (ast->type() == TypedASTTag::Identifier) {
 		auto as_id = static_cast<TypedAST::Identifier*>(ast);
 		auto decl = as_id->m_declaration;
 		auto value =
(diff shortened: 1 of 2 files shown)


src/typecheck.cpp
```diff
@@ -87,22 +87,13 @@ void typecheck(TypedAST::Identifier* ast, TypeChecker& tc) {
...
-	} else if (meta_type == tc.meta_typefunc()) {
-		assert(0 && "Accessed a name of a typefunc (not implemented)");
-		// we are a type function.
-		// TODO: not too sure what needs to be done...
-	} else {
-		assert(0 && "Accessed a name with unknown metatype (not implemented)");
-		// meta type var
-		// TODO: not too sure what needs to be done...
-	}
+	assert(meta_type == tc.meta_value());
```
lushoBarlett, line 91: Okaaaaay, now this makes a lot more sense. I guess I was confused as to what the purpose and responsibility differences were between metacheck, ct_eval and typecheck.

src/ct_eval.cpp
```diff
@@ -33,6 +33,45 @@ TypedAST::FunctionLiteral* ct_eval(
...
+	assert(ast);
+	assert(ast->m_declaration);
+
+	// TODO: These two lines are fishy; I copied them from
+	// elsewhere, but I don't think we need to go through
+	// the declaration to get the metatype: this is already
+	// done for us by metacheck().
+	MetaTypeId meta_type =
+	    tc.m_core.m_meta_core.find(ast->m_declaration->m_meta_type);
+	ast->m_meta_type = meta_type;
```
lushoBarlett, line 53: All of this is already done in metacheck, yes. But not quite. In metacheck there is no find. The identifier just points to the declaration's meta type... I guess this lines can still be deleted?

lushoBarlett (approved):
Looks like it will do the trick. Now in typecheck, all info on types, I can get from Handles.

src/ct_eval.cpp
SebastianMestre: That makes sense to me. I will merge it as-is, and look into it later today

PR #178, opened 2020-10-23
use the correct index when instanciating a type

SebastianMestre:
A small bug was introduced yesterday, where we used the type func id as an index into the type function data array, instead of the type function header array. This patch fixes that.

A more permanent solution would be to use more strongly typed indices with accessor functions, which would catch these mistakes at compile.

I'll work on that later today or tomorrow.


lushoBarlett (approved):
Thank you for catching this, seems like it was on `new_term` as well

PR #179, opened 2020-10-25
remove superfluous includes

SebastianMestre:
I shaved like a second or two off of the build time by removing superfluous includes.

A large piece of time (~30%) is spent instanciating `std::vector` for all the different types of `AST` and `TypedAST` nodes. (because we use `std::vector` in `ChunkedArray`, which is itself the core piece of `NodeAllocator`)... I have some ideas to replace it with something that's less compile-time intensive, but that's better left for tomorrow.


lushoBarlett (approved):
A little sad that this is only one second. Perhaps changing `ChunkedArray` will yield greater speedups

PR #180, opened 2020-10-25
Implement a new block allocator

SebastianMestre:
The current block allocator (NodeAllocator) does its job of providing faster runtime performance, but it is abysmal for compilation time (it currently takes about 30% of the build time).

In this PR, I want to start a discussion about allocator design, and the associated trade-offs. Hopefully, leading to a new allocator we can use.


SebastianMestre:
Here is one of the designs I have in mind:

A block allocator owns a chunk of memory (aka a block), which it partitions into evenly sized pieces (aka slots).

```
[=BlockAllocator======]
[  int slot_count_per_block ]
[ int block_size_in_bytes]
[  int slot_size_in_bytes]
[          uin8_t* block] -----
                              |
```

SebastianMestre:
I tried replacing ast and typed ast allocators, compilation got ~2x faster, with a 20% runtime performance degradation on a particular benchmark.

SebastianMestre:
After some tuning, I managed to get it pretty much just as fast as it used to be, but the build now takes under 9 seconds on my machine

SebastianMestre:
i noticed that the main slowness of this new allocation comes from page faults when using previously-unused parts of the allocator. I don't know if there is a way to load those pages preemptively without incurring page fault. I will look into it.

SebastianMestre:
I thought of a different scheme. Since we are storing only AST nodes, or only TypedAST nodes in each allocator, we could rely on virtual destructors being present. This way, we wouldn't have the 8 byte overhead on every object

SebastianMestre:
I've implemented the destructor for automatic allocators. It's really really slow.

lushoBarlett:
I don't think we destroy that much, I mean, we do but at the very end. The destructor you store just seems like you are "pre-loading" the specific destructor function, so no v-table is required. The relative speed of both approaches seems to depend on if we would like to traverse the blocks in a linear fashion?

If more stuff is in there, then you have load more times for the same data we want to read. But I think we don't do that, we traverse a tree accessed by pointers, that is already as slow as the thing I mentioned could potentially be a small issue.

My conclusion is: do store the destructors

SebastianMestre:
> I don't think we destroy that much, I mean, we do but at the very end. The destructor you store just seems like you are "pre-loading" the specific destructor function, so no v-table is required.

Well, yes, I think so too, but I can still come up with some counter-arguments

 - storing the function pointer is also a form of vtable. So, we still get an indirect call.
 - omitting the extra pointer lets us fit more objects per allocation.
 - (as you mentioned) omitting the extra pointer lets us fit more objects per cache line, and do fewer loads.
 - calling the virtual destructor on all objects means all the vtables get cached, so it could still be fast

to expand on some of these points:

by doing virtual destructor calls, we get (assuming we don't need to cache things twice):
 - 2T extra cache misses, where T is the amount of different types in the allocator. (one for each vtable, one for each destructor)
 - 0.2N fewer cache misses, where N is the amount of objects in the allocator. (because each slot is 8 bytes smaller)

Here is my work, in case I made a mistake: every 15 cache lines (`15 * 64 bytes = 960 bytes`), we can fit 20 slots (`960 bytes / 48 bytes per slot = 20 slots`) using the explicit pointer scheme, and 24 slots (`960B / 40B = 24`) using the virtual destructor scheme.

`(24 - 20) / 20 = 4 / 20 = 2 / 10 = 20%`

> The relative speed of both approaches seems to depend on if we would like to traverse the blocks in a linear fashion?
>
> If more stuff is in there, then you have load more times for the same data we want to read. But I think we don't do that, we traverse a tree accessed by pointers, that is already as slow as the thing I mentioned could potentially be a small issue.

You would be right, but destroying the allocator is essentially a linear scan, which might be significant for overall runtime.

> My conclusion is: do store the destructors

I don't disagree (and that was what I went with initially) but I think it would be best to try out both approaches before making a choice.

SebastianMestre:
Surprisingly enough, the vtable based allocator does seem slightly faster than the one that stores the destructors explicitly.

the virtual destructor one is in the high 0.017s to low 0.018s, whereas the function pointer one is in the low-to-middle 0.018s

It is a very small margin, but it seems consistent. (I don't discard the possibility that it's just noise, but I don't feel like doing more rigorous testing right now)

SebastianMestre:
We have a big object allocator now. It's really stupid. It just puts objects in their own allocation, and then stores a pointer to those objects in a block allocator.

SebastianMestre:
I think this is pretty good. I'm not removing the AutomaticBlockAllocator yet, in case you want to measure which one is faster

lushoBarlett (approved):
I glossed over it, but I did not study it carefully. I'm gonna approve it so you can merge it if you want tho.

PR #181, opened 2020-10-28
Structs

SebastianMestre:
Implements struct in the runtime


SebastianMestre:
I wanted to make it prettier but, whatever, this is good enough

lushoBarlett (approved):
I don't like how we store the type checker in the runtime environment... but I guess it works for now. My idea was to store the required data in the Handle nodes.

If you don't want to do it, I'll consider doing it myself later.

Also, test?

PR #186, opened 2020-11-03
Bugfix capture ordering

### LINKED ISSUE #184: Example code prints 'String "!WorldHello, "' instead of 'String "Hello World!'
Example code prints 'String "!WorldHello, "' instead of 'String "Hello World!'. The code seems to be implemented to print the correct version, is this a bug in the interpreter?

SebastianMestre:
Fixes #184 

We were pushing captures onto the stack in an arbitrary order, and not the one generated by `compute_offsets`.

### DIFF
--- a/src/interpreter/eval.cpp
+++ b/src/interpreter/eval.cpp
@@ -150,10 +150,17 @@ void eval_call_function(
 	// TODO: error handling ?
 	assert(callee->m_def->m_args.size() == arg_count);
 
+	// TODO: This is a big hack. pushing nullptr into the
+	// stack should actually never happen.
+	for (int i = callee->m_captures.size(); i--;)
+		e.push(nullptr);
+
 	for (auto& kv : callee->m_captures) {
 		assert(kv.second);
 		assert(kv.second->type() == ValueTag::Reference);
-		e.push(kv.second);
+		auto capture_value = unboxed(kv.second);
+		auto offset = callee->m_def->m_captures[kv.first].inner_frame_offset;
+		e.m_stack[e.m_frame_ptr + offset] = kv.second;
 	}
 
 	assert(callee->m_def->m_body->type() == TypedASTTag::Block);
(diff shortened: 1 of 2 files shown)


src/interpreter/eval.cpp
```diff
@@ -150,10 +150,17 @@ void eval_call_function(
...
+	for (int i = callee->m_captures.size(); i--;)
+		e.push(nullptr);
+
 	for (auto& kv : callee->m_captures) {
 		assert(kv.second);
 		assert(kv.second->type() == ValueTag::Reference);
-		e.push(kv.second);
+		auto capture_value = unboxed(kv.second);
+		auto offset = callee->m_def->m_captures[kv.first].inner_frame_offset;
+		e.m_stack[e.m_frame_ptr + offset] = kv.second;
```
lushoBarlett, line 163: Ah yes, very beautiful code. You are a true genie. Thanks.

src/test/main.cpp
```diff
@@ -455,6 +455,18 @@ void interpreter_tests(Test::Tester& tests) {
...
+
+	tests.add_test(std::make_unique<Test::InterpreterTestSet>(
+	    R"(
+			cat := fn(a,c,d) => fn(b) =>
+				a + b + c + d;
+
+			__invoke := fn() =>
+				cat("A","C","D")("B");
+		)",
+	    Testers {+[](Interpreter::Environment& env) -> ExitStatusTag {
```
lushoBarlett, line 467: Cool test, very helpful. Maybe we could use RapidTest next time?

lushoBarlett (approved):
Yeah! Great work!

src/interpreter/eval.cpp
SebastianMestre: thx

src/test/main.cpp
SebastianMestre: :^)

PR #188, opened 2020-11-06
blow up when encountering a declaration

SebastianMestre:
We used to ignore re-declarations. Now we have an assertion to forbid them. Not a nice thing to do, but way better than failing silently.


lushoBarlett (approved):
Pretty necessary, I agree

PR #189, opened 2020-11-06
Refactor Runtime Environment

SebastianMestre:
This class had too much stuff going on. I moved most of it to the more aptly named `Interpreter::Interpreter`

Now, `Interpreter::Environment` only handles the stack, so, maybe, a change of name is in order.


lushoBarlett (approved):
I see how `Environment` had many things going for it. But now it's just a stack handler, and thus I believe its name is obsolete and should be changed to something else. Plus, `env.m_env` is really bad...

So there's actually more than one reason why the name should be changed.

SebastianMestre:
@lushoBarlett 

> ... is obsolete and should be changed ... is really bad ... more than one reason why [it] should be changed

While I agree, I would prefer if you held off on this type of remarks if they are not accompanied with a suggested solution or alternative.

SebastianMestre:
I'll rename `Interpreter::Environment` at another time.

PR #190, opened 2020-11-06
Type hints

lushoBarlett:
Implements declaration type hints, also allows parsing some type keywords as identifiers (for type terms) and declares `array` as a possible type in typechecker.

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -221,8 +221,8 @@ MonoId mono_type_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc){
 
 TypedAST::Declaration* ct_eval(
     TypedAST::Declaration* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
-	if (ast->m_type)
-		ast->m_type = ct_eval(ast->m_type, tc, alloc);
+	if (ast->m_type_hint)
+		ast->m_type_hint = ct_eval(ast->m_type_hint, tc, alloc);
 
 	ast->m_value = ct_eval(ast->m_value, tc, alloc);
 	return ast;
@@ -264,8 +264,8 @@ TypedAST::DeclarationList* ct_eval(
 			MonoId mt = mono_type_from_ast(handle->m_syntax, tc);
 			tc.m_core.m_mono_core.unify(mt, handle->m_value);
 		} else {
-			if (decl.m_type)
-				decl.m_type = ct_eval(decl.m_type, tc, alloc);
+			if (decl.m_type_hint)
+				decl.m_type_hint = ct_eval(decl.m_type_hint, tc, alloc);
 			decl.m_value = ct_eval(decl.m_value, tc, alloc);
 		}
 	}
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -221,6 +221,9 @@ MonoId mono_type_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc){
 
 TypedAST::Declaration* ct_eval(
     TypedAST::Declaration* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	if (ast->m_type)
+		ast->m_type = ct_eval(ast->m_type, tc, alloc);
+
 	ast->m_value = ct_eval(ast->m_value, tc, alloc);
 	return ast;
 }
@@ -237,7 +240,7 @@ TypedAST::DeclarationList* ct_eval(
 			handle->m_value = tc.m_core.m_tf_core.new_var();
 			handle->m_syntax = decl.m_value;
 			decl.m_value = handle;
-		} else if(meta_type == tc.meta_monotype()) {
+		} else if (meta_type == tc.meta_monotype()) {
 			auto handle = alloc.make<TypedAST::MonoTypeHandle>();
 			handle->m_value = tc.new_var(); // should it be hidden?
 			handle->m_syntax = decl.m_value;
@@ -247,9 +250,11 @@ TypedAST::DeclarationList* ct_eval(
 
 	for (auto& decl : ast->m_declarations) {
 		int meta_type = tc.m_core.m_meta_core.find(decl.m_meta_type);
+		// TODO: ban type hints from type function and mono type declarations?
 		if (meta_type == tc.meta_typefunc()) {
 			auto handle =
 			    static_cast<TypedAST::TypeFunctionHandle*>(decl.m_value);
+
 			TypeFunctionId tf = type_func_from_ast(handle->m_syntax, tc);
 			tc.m_core.m_tf_core.unify(tf, handle->m_value);
 		} else if (meta_type == tc.meta_monotype()) {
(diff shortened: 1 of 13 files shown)


lushoBarlett:
While implementing this I realized there's duplication in `Declaration` and `DeclarationList` some times, and that caused me problems, because I had written code for one of the functions and not for the other.

src/metacheck.cpp
```diff
@@ -144,6 +144,12 @@ void metacheck(TypedAST::ReturnStatement* ast, TypeChecker& tc) {
 
 void metacheck(TypedAST::Declaration* ast, TypeChecker& tc) {
 	ast->m_meta_type = tc.new_meta_var();
+
+	if (ast->m_type) {
+		metacheck(ast->m_type, tc);
+		assert(ast->m_type->m_meta_type != tc.meta_value() && "Type hint must not be a value");
```
SebastianMestre: Typehint must be a monotype, actually

src/metacheck.cpp
```diff
@@ -152,8 +158,13 @@ void metacheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
 	for (auto& decl : ast->m_declarations)
 		decl.m_meta_type = tc.new_meta_var();
 
-	for (auto& decl : ast->m_declarations)
+	for (auto& decl : ast->m_declarations) {
+		if (decl.m_type) {
+			metacheck(decl.m_type, tc);
+			assert(decl.m_type->m_meta_type != tc.meta_value() && "Type hint must not be a value");
```
SebastianMestre: Typehint must be a monotype

src/typecheck.cpp
```diff
@@ -57,6 +59,17 @@ void typecheck(TypedAST::Declaration* ast, TypeChecker& tc) {
 
 	ast->m_value_type = tc.new_var();
 
+	if (ast->m_type) {
+		if (ast->m_type->type() == TypedASTTag::MonoTypeHandle) {
```
SebastianMestre: typehints must be monotypes. This should be an assertion, not a check.

src/typecheck.cpp
```diff
@@ -57,6 +59,17 @@ void typecheck(TypedAST::Declaration* ast, TypeChecker& tc) {
...
+			auto handle = static_cast<TypedAST::MonoTypeHandle*>(ast->m_type);
+			tc.m_core.m_mono_core.unify(ast->m_value_type, handle->m_value);
+		}
+		else {
+			auto handle = static_cast<TypedAST::TypeFunctionHandle*>(ast->m_type);
+			tc.m_core.m_tf_core.unify(ast->m_value_type, handle->m_value);
```
SebastianMestre: Pretty sure this is a bug. `ast->m_value_type` must always be a monotype.

src/typecheck.cpp
```diff
@@ -336,6 +349,17 @@ void typecheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
 		for (int u : verts) {
 			auto decl = index_to_decl[u];
 
+			if (decl->m_type) {
+				if (decl->m_type->type() == TypedASTTag::MonoTypeHandle) {
```
SebastianMestre: assert, don't check

src/typecheck.cpp
```diff
@@ -336,6 +349,17 @@ void typecheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
...
+					auto handle = static_cast<TypedAST::MonoTypeHandle*>(decl->m_type);
+					tc.m_core.m_mono_core.unify(decl->m_value_type, handle->m_value);
+				}
+				else {
+					auto handle = static_cast<TypedAST::TypeFunctionHandle*>(decl->m_type);
+					tc.m_core.m_tf_core.unify(decl->m_value_type, handle->m_value);
```
SebastianMestre: this case is a bug

src/ct_eval.cpp
```diff
@@ -247,9 +250,11 @@ TypedAST::DeclarationList* ct_eval(
 
 	for (auto& decl : ast->m_declarations) {
 		int meta_type = tc.m_core.m_meta_core.find(decl.m_meta_type);
+		// TODO: ban type hints from type function and mono type declarations?
```
SebastianMestre, line 253: Yes. That's probably wise.

SebastianMestre (changes requested):
I don't think we should allow type functions in hints. Only monotypes. We can support that sugar in the future.

src/metacheck.cpp
lushoBarlett: Good to know, I wasn't sure myself.

lushoBarlett:
> I don't think we should allow type functions in hints; only monotypes. We can add support for more sugar in the future.

That does simplify things a lot, I wasn't sure so I included the type function case as well, will delete it now tho.

src/ct_eval.cpp
```diff
@@ -250,14 +250,15 @@ TypedAST::DeclarationList* ct_eval(
 
 	for (auto& decl : ast->m_declarations) {
 		int meta_type = tc.m_core.m_meta_core.find(decl.m_meta_type);
-		// TODO: ban type hints from type function and mono type declarations?
 		if (meta_type == tc.meta_typefunc()) {
+			assert(!decl.m_type_hint && "type hint not allowed in type function declaration");
```
SebastianMestre, line 254: I feel like this check should be in metacheck, and not here. Thoughts?

src/metacheck.cpp
```diff
@@ -147,7 +147,7 @@ void metacheck(TypedAST::Declaration* ast, TypeChecker& tc) {
 
 	if (ast->m_type_hint) {
 		metacheck(ast->m_type_hint, tc);
-		assert(ast->m_type_hint->m_meta_type != tc.meta_value() && "Type hint must not be a value");
+		assert(ast->m_type_hint->m_meta_type == tc.meta_monotype() && "Type hint must be a monotype");
```
SebastianMestre, line 150: I'm not sure... maybe we should unify instead of check?

src/metacheck.cpp
```diff
@@ -161,7 +161,7 @@ void metacheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
 	for (auto& decl : ast->m_declarations) {
 		if (decl.m_type_hint) {
 			metacheck(decl.m_type_hint, tc);
-			assert(decl.m_type_hint->m_meta_type != tc.meta_value() && "Type hint must not be a value");
+			assert(decl.m_type_hint->m_meta_type == tc.meta_monotype() && "Type hint must be a monotype");
```
SebastianMestre, line 164: Maybe unify instead of check? idk

SebastianMestre:
You mentioned how its still accepted things like `x : int = 10;`, even though you removed the special handling for bare type functions.

That seemed really weird, but it actually makes a lot of sense. It's because of a quirk in the parser: when we already know we are parsing a type term, we implicitly wrap identifiers in type terms with no arguments, which means we get some nice syntactic sugar for free!

SebastianMestre (approved):
Looks good! Seems to be working quite well!

I have a few doubts, but they are more about tidiness and robustness, so they are not really urgent.

src/ct_eval.cpp
lushoBarlett: I thought so as well, but it seemed like much less work to put them there hahaha. Perhaps we will fix that when we clean up.

src/metacheck.cpp
lushoBarlett: Fair point. I still like to have an assert there for us when we debug and test.

PR #195, opened 2020-11-12
add compile time value header

SebastianMestre:
Right now, we have, essentially, a compile time interpreter for type expressions. This walks ASTs recursively and replaces some expressions with a handle to their corresponding value.

The type it uses to represent these things is `TypedAST::TypedAST`. To represent type handles, we have some `TypedAST::xxxHandle` types. I feel like this abstraction is very messy. Maybe we would benefit from rethinking how this works?

My idea is roughly as follows:

have a base type, called `CompileTimeValue`, or something like that.

Have a few different subtypes, each one responsible for storing

 - An expression (i.e. a `TypedAST` node)
 - A type function handle
 - A mono type handle

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ COMMON_TARGETS := \
 	utils/string_view \
 	ast \
 	compile_time_environment \
+	compile_time_value \
 	compute_offsets \
 	ct_eval \
 	desugar \
--- a/src/compile_time_value.cpp
+++ b/src/compile_time_value.cpp
@@ -1,0 +1,1 @@
+#include "compile_time_value.hpp"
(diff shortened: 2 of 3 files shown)


SebastianMestre:
After letting the though sit in my mind for a bit longer, this is probably worse than what we have currently.

A name change for `TypedAST` is probably enough.

```c++
struct CTValue;

struct MonoHandle final : CTValue;
struct FuncHandle final : CTValue;

struct Declaration final : CTValue;
struct CallExpression final : CTValue;
...
```

*Maaaybe* we could have a two level inheritance hierarchy?

```c++
struct CTValue;

struct MonoHandle final : CTValue;
struct FuncHandle final : CTValue;

struct TypedAST : CTValue;
struct Declaration final : TypedAST;
struct CallExpression final : TypedAST;
...
```

I'm not sure how helpful that'd be though.

PR #197, opened 2020-11-12
Cleanup tc

SebastianMestre:
The TC was built up in a very non-deliberate way. This led to quite a bit of mess and accidental complexity, some of it I attempt to address in this PR.

### DIFF
--- a/src/typesystem.cpp
+++ b/src/typesystem.cpp
@@ -19,6 +19,12 @@ TypeSystemCore::TypeSystemCore() {
 		TypeFunctionData& b_data = m_type_functions[fb];
 
 		if (a_data.is_dummy) {
+
+			// TODO: handle variadic arguments properly
+			// for example: we should be able to unify a dummy with 5
+			// arguments with a non dummy with variadic arguments
+			assert(a_data.argument_count == b_data.argument_count);
+
 			for (auto& kv_a : a_data.structure) {
 				auto kv_b = b_data.structure.find(kv_a.first);
 
@@ -50,7 +56,6 @@ TypeSystemCore::TypeSystemCore() {
 		} else {
 			assert(0 and "unifying two different known type functions");
 		}
-
 	};
 
 	m_mono_core.unify_function = [this](Unification::Core& core, int a, int b) {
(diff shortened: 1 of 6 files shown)


SebastianMestre:
about recursive typefuncs:

They complicate things. In particular, it makes it possible to invoke a typefunc before we know how many arguments it takes. This means that we can no longer just check the argument count when creating a typefunc.

Instead, we need to do something like the Hindley-Milner app rule, where we create a dummy typefunc that takes as many arguments as it is passed in the type term, and the unify it with the typefunc. If we pass a bad argument count, it could fail immediately (if we already know the argument count of the typefunc), or later (when we learn the argument count of the typefunc).

This makes some things in the implementation harder to think about, but I think it's worth it.

src/typesystem.cpp
```diff
@@ -19,6 +19,16 @@ TypeSystemCore::TypeSystemCore() {
...
 		if (a_data.is_dummy) {
+
+			int new_argument_count = [&] {
+				if (a_data.argument_count == b_data.argument_count)
+					return a_data.argument_count;
+				else if (b_data.is_dummy || b_data.argument_count == -1)
+					return -1;
+				else
+					assert(0 && "Incompatible typefunc argument counts");
+			}();
```
lushoBarlett, line 30: Wow this is unnecessary... but fancy. Perhaps even clearer. Hope it compiles to the same thing!

src/typesystem.cpp
```diff
@@ -100,11 +110,20 @@ TypeSystemCore::TypeSystemCore() {
...
+		// that it's a dummy of unknown characteristics
+
+		// TODO: add some APIs to make this less jarring
+		TypeFunctionId dummy_tf =
+		    new_type_function(TypeFunctionTag::Builtin, {}, {}, true);
+
+		int dummy_tf_data_id = m_tf_core.find_function(dummy_tf);
+		m_type_functions[dummy_tf_data_id].argument_count = args.size();
+
+		m_tf_core.unify(tf, dummy_tf);
```
lushoBarlett, line 125: So what this is saying is: we may not know what our type function actually is yet, but we do know the arguments we are instanciating this with right now. So we create a new dummy type function with that information so as to unify with other stuff and check for errors that way. Makes sense.

lushoBarlett (approved):
Thank you for addressing code duplication in typecheck! I think that was the worst case, because it was pretty clear. But I think there is some of it in other phases as well.

src/typesystem.cpp
SebastianMestre: exactly

PR #198, opened 2020-11-12
lex keywords by lexing identifiers then checking against a table

SebastianMestre:
Someone suggested this on Reddit. I really liked the idea: it's simpler, shorter and probably not a whole lot slower. In fact, it can probably be made faster than the switch-case thing with little effort.


lushoBarlett (approved):
It's probably still quite fast because of `InternedString`. I do like it a lot, and since the `Lexer` is not a primary speed concern I'm sure it's fine

PR #199, opened 2020-11-13
Union integration

lushoBarlett:
This pr finishes the implementation of unions. It will also attempt to adapt the `RecordAccessExpression` as to allow this:
```rust
x := my_union.field { value };
```

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -173,7 +173,6 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 	} else if (ast->type() == TypedASTTag::UnionExpression) {
 		auto as_ue = static_cast<TypedAST::UnionExpression*>(ast);
 
-		std::vector<InternedString> constructors;
 		std::unordered_map<InternedString, MonoId> structure;
 		int constructor_count = as_ue->m_constructors.size();
 		for (int i = 0; i < constructor_count; ++i){
@@ -181,11 +180,10 @@ TypeFunctionId type_func_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc) {
 			InternedString name = as_ue->m_constructors[i].text();
 			assert(!structure.count(name));
 			structure[name] = mono;
-			constructors.push_back(name);
 		}
 
 		TypeFunctionId result = tc.m_core.new_type_function(
-		    TypeFunctionTag::Sum, std::move(constructors), std::move(structure));
+		    TypeFunctionTag::Sum, {}, std::move(structure));
 
 		return result;
 	} else if (ast->type() == TypedASTTag::StructExpression) {
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -247,7 +247,19 @@ TypedAST::Constructor* constructor_from_ast(
 		assert(ast->type() == TypedASTTag::AccessExpression);
 
 		auto access = static_cast<TypedAST::AccessExpression*>(ast);
-		constructor->m_mono = mono_type_from_ast(access->m_object, tc);
+
+		// dummy with one constructor, the one used
+		std::unordered_map<InternedString, MonoId> structure;
+		structure[access->m_member->m_text] = tc.new_var();
+		TypeFunctionId dummy_tf = tc.m_core.new_type_function(
+		    TypeFunctionTag::Sum, {}, std::move(structure), true);
+
+		MonoId monotype = mono_type_from_ast(access->m_object, tc);
+		TypeFunctionId tf = tc.m_core.m_mono_core.find_function(monotype);
+
+		tc.m_core.m_tf_core.unify(dummy_tf, tf);
+
+		constructor->m_mono = monotype;
 		constructor->m_id = access->m_member;
 	} else {
 		assert(0 && "wrong constructor metatype");
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -96,9 +96,9 @@ TypedAST::TernaryExpression* ct_eval(
 	return ast;
 }
 
-TypedAST::RecordAccessExpression* ct_eval(
-    TypedAST::RecordAccessExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
-	ast->m_record = ct_eval(ast->m_record, tc, alloc);
+TypedAST::AccessExpression* ct_eval(
+    TypedAST::AccessExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	ast->m_object = ct_eval(ast->m_object, tc, alloc);
 	return ast;
 }
 
@@ -243,11 +243,11 @@ TypedAST::Constructor* constructor_from_ast(
 		    ct_eval(ast, tc, alloc));
 		constructor->m_mono = mono_handle;
 	} else if (meta == tc.meta_constructor()) {
-		assert(ast->type() == TypedASTTag::RecordAccessExpression);
+		assert(ast->type() == TypedASTTag::AccessExpression);
 
-		auto access = static_cast<TypedAST::RecordAccessExpression*>(ast);
+		auto access = static_cast<TypedAST::AccessExpression*>(ast);
 		auto mono_handle = static_cast<TypedAST::MonoTypeHandle*>(
-		    ct_eval(access->m_record, tc, alloc));
+		    ct_eval(access->m_object, tc, alloc));
 
 		constructor->m_mono = mono_handle;
 		constructor->m_id = access->m_member;
@@ -346,7 +346,7 @@ TypedAST::TypedAST* ct_eval(
 		DISPATCH(CallExpression);
 		DISPATCH(IndexExpression);
 		DISPATCH(TernaryExpression);
-		DISPATCH(RecordAccessExpression);
+		DISPATCH(AccessExpression);
 		DISPATCH(ConstructorExpression);
 
 		DISPATCH(Block);
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -96,8 +96,13 @@ TypedAST::TernaryExpression* ct_eval(
 	return ast;
 }
 
-TypedAST::AccessExpression* ct_eval(
+TypedAST::TypedAST* ct_eval(
     TypedAST::AccessExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	MetaTypeId metatype = tc.m_core.m_meta_core.find(ast->m_meta_type);
+	// TODO: support vars
+	if (metatype == tc.meta_constructor())
+		return constructor_from_ast(ast, tc, alloc);
+
 	ast->m_object = ct_eval(ast->m_object, tc, alloc);
 	return ast;
 }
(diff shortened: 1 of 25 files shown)


SebastianMestre:
You'll probably have to do a lot of work in metacheck and ct_eval to get this to work.

note how

    e is a union monotype => e.id is a constructor
    e is a value => e.id is a value

lushoBarlett:
I was thinking that the way the constructor expression is parsed is:
```
     e.id         {...}
constructor args
```
So in particular I need to know that `e.id` is a valid thing to be constructed, i.e a monotype. So what I was thinking was that an `AccessExpression` on a union type function returns the monotype of that accessed member.

I get what you mean, that the returned type of that construction is not actually the monotype of the member, but instead a _union-ised_ monotype. So I have to develop a nice way to make that happen. But that is the plan so far.

SebastianMestre:
To clarify, the monotype of `my_union(<>).id { ... }` should be `my_union(<>)`

lushoBarlett:
Ah, now I understand, well I'll have to think about this for a bit.

lushoBarlett:
I'm having trouble with this thing, and I don't think the compile time record trick is a solution to it all. The current way we think about constructors is that the left thing is a monotype and the right thing are it's runtime arguments. So it returns a value of that monotype.

But there's another problem in my opinion. Think about a union monotype called `um`, from there, you need to access a member of it, constructor `id`. Each constructor is unique by itself, and the monotype it has is irrelevant. In the end, whatever approach you choose, the resulting runtime value of a construction is of type `um`, but it needs to remember that it was constructed with `id`.

This made me think of things:

1. Special union runtime value, whose type is `um`, but it has a `constructed_with` and it has an internal value. Very cheap and straight forward solution, thus maybe not the best, but the eventual `match` function suddenly has something to work with!

2. Special function in the type realm, that takes the constructor's identification (name or otherwise) and builds a new monotype from it that somehow stores information of `um` and `id`. Thus the constructor expression is valid because this is a monotype. I don't like this at all because it taints the typesystem with a very special case, which is worse in my opinion than doing so in the runtime values.

3. Like you said, make the constructor expression a bit more interesting, maybe allowing a more general thing called "constructable" and the rules by which it is constructed are to be specified (by us). I have no clue what the idea here would be tho, I don't see this as good and this could still cause problems with runtime values.

Approach 1 is pretty solid, but we can kiss `{...}` syntax bye bye probably. 2 seems messy and tedious, but it allows `{...}`, also probably `match` still works fine with this too. 3 I just don't what to do.

I need an opinion, if none is provided (or if it's not strong enough) I think I will just mess around with the first approach.

SebastianMestre:
There are two separate issues here

1. The runtime representation of a union value
1. The compile time representation of a union and its constructors

These can be tackled in isolation from each other.

# The run time representation
This one is quite simple: you need to store a value, and some way to know which variant the union is holding. (Note that we tackle this very problem all over the compiler, in the ASTs)

The simplest way would be to use a struct that has a pointer to a value and an integer that can be used as a tag.

# The compile time representation (the interesting part)

## Solution 1

My first idea is a compile time record, where

```
t := union {
  a : int(<>);
  b : string(<>);
}
```

Evaluates to something with this shape:

```
struct {
  a : int(<>) -> t(<>);
  b : string(<>) -> t(<>);
}
```

And all those functions (`a` and `b`) do the right thing at runtime. (i.e. create  a pointer-integer pair with the right values, assuming my idea for a runtime representation)

Note thay we probably want to implement compile time records regardless of whether it's useful here, in order to support things like modules and such.

## Solution 2

Have constructors be their own thing.

Having construct-expressions act on monotypes doesnt really make sense for things that arent records or tuples (what would `int(<>){...}` or `my_union(<>){ ... }` mean?), so this is probably a good idea, independently of how useful it is in implementing unions.

Now, construct-expressions can act on constructors, and things seem to work out quite nicely (I haven't thought about this too deeply). In particular, if accessing a member of a union monotype gives a constructor (that is known at compile time), then we can just invoke it at runtime and it should do the right thing.

# closing notes

There are no special cases in the runtime. A tag is just inherently needed to implement this functionality.

There is no need to create more monotypes. The monotype of instanciating a union type is always that union type, regardless of what it holds. (Think about what Haskell does. The types of `Left expr` and `Right expr` are both `Either a`)

lushoBarlett:
As a sidenote, I thought that `int(<>) { 3 }` is just 3. Just like `array { ... }` but this is implemented as a special case, and I thought that could be changed if constructors are implemented more generally

SebastianMestre:
> But having constructors be "their own thing" I still don't get what you mean exactly.

They are not monotypes. If they are not functions, then they must be something else.

lushoBarlett:
Yea. After giving it some thought, I have a few ideas. So buckle up for what probably is the wrong approach but it may just be good enough!

lushoBarlett:
Not the prettiest thing in the world, but it does seem to work!

SebastianMestre:
Hey! Can you give me a high level description of what the changes are, and what the "big idea" is here? just so that I don't jump in blind into the review. thanks!

lushoBarlett:
Core changes here:
- added a new metatype, constructor
- changed RecordAccessExpression to allow things like monotype.member
- created a Union value, UnionConstructor (similar to StructConstructor) and Constructor TypedAST node.

The Constructor is the big part I think, together with the RecordAccessExpression change. The Constructor node holds a monotype handle always, and a member only if it is a union accessing. So it's basically ad-hoc.

lushoBarlett:
I should note, again, everything's really messy and ad-hoc. I basically invented some stuff to make the thing work, analyzing only the case I'm interested in. Looking back there's probably unneeded things and good abstractions that you might discover.

But it does work and I'm happy about that.

src/typed_ast.hpp
```diff
@@ -317,4 +317,14 @@ struct MonoTypeHandle : public TypedAST {
 	    : TypedAST {TypedASTTag::MonoTypeHandle} {}
 };
 
+struct Constructor : public TypedAST {
+	MonoTypeHandle* m_mono;
```
SebastianMestre, line 321: Why not use a `MonoId` instead of a node here?

src/ct_eval.cpp
```diff
@@ -94,15 +96,15 @@ TypedAST::TernaryExpression* ct_eval(
...
 }
 
-TypedAST::RecordAccessExpression* ct_eval(
-    TypedAST::RecordAccessExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
-	ast->m_record = ct_eval(ast->m_record, tc, alloc);
+TypedAST::AccessExpression* ct_eval(
+    TypedAST::AccessExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	ast->m_object = ct_eval(ast->m_object, tc, alloc);
 	return ast;
 }
```
SebastianMestre, line 103: It would be cool if this returned a `TypedAST::Constructor*` when the metatype is `constructor`

SebastianMestre (commented):
I'm sorry, I can't leave a full review right now, and I like looking at things before accepting, so I am leaving this open for now. I will look at this tomorrow. For now, I left you some points to think about.

The overall structure looks good, but I do wonder if the `TypedAST::Constructor` idea could be fleshed out a bit more. Thanks!

src/typed_ast.hpp
SebastianMestre: Also, this `m_mono` probably be defaulted to `nullptr`, just in case

src/ct_eval.cpp
SebastianMestre: I feel like this is more consistent with the idea that `ct_eval` returns the compile time value of an expression

src/ct_eval.cpp
```diff
@@ -211,6 +232,32 @@ MonoId mono_type_from_ast(TypedAST::TypedAST* ast, TypeChecker& tc){
...
+		constructor->m_mono = mono_handle;
+	} else if (meta == tc.meta_constructor()) {
+		assert(ast->type() == TypedASTTag::AccessExpression);
+
+		auto access = static_cast<TypedAST::AccessExpression*>(ast);
+		auto mono_handle = static_cast<TypedAST::MonoTypeHandle*>(
+		    ct_eval(access->m_object, tc, alloc));
+
+		constructor->m_mono = mono_handle;
+		constructor->m_id = access->m_member;
```
SebastianMestre, line 253: What would happen if we haven't yet evaluated the type func when we reach this point?

example:

```
x := my_union(<>).a { 10 };
my_union := union {
  a : int(<>);
};
```

SebastianMestre: If this is wrong as I think it is, the fix would be to use unification, kinda like the thing we do when we access a member of a record, but we potentially don't know the record yet.

src/ct_eval.cpp
lushoBarlett: I did not think about this (I always forget about that). I will add a test and try to make it work then, so don't worry about merging for now

src/ct_eval.cpp
lushoBarlett: I did not think about this case. I will add a test and make it work

src/ct_eval.cpp
lushoBarlett: I will look into it.

src/typed_ast.hpp
lushoBarlett: Good point, I can't remember what lead me to store a handle, but all the info I need requires only the id.

src/ct_eval.cpp
lushoBarlett: Doing the example you suggested seems to work. I think it works because `match_identifiers` takes care of it, so by the time we compile, the constructor creator calls the mono type creator, which calls the type function creator that does the rest.

lushoBarlett:
Okay, so, that test is quite good for trying the recursive part I think. Turns out it works. I understand what you are saying about unifying to a dummy, and why you are saying it, so I'll probably do it anyway. But it's nice to see something working.

src/ct_eval.cpp
SebastianMestre: Oh I'm surprised. I'll give your patch a more thorough reading in a bit then

SebastianMestre: Ah yes, recursive typefuncs will probably expose a bunch of hidden assumptions throughout the codebase

SebastianMestre:
Hey, I downloaded it, tested it out, and it worked flawlessly.

I also implemented print, because it was nice to have while testing things.

```c++
void print(Union* m, int d) {
	print_spaces(d);
	std::cout << value_string[int(m->type())] << " " << m->m_constructor << '\n';
	print(m->m_inner_value, d);
}
```

src/ct_eval.cpp
SebastianMestre: I don't understand why it works yet, but it looks like it does... huh

SebastianMestre (commented):
I found a few error cases that aren't handled.

this one breaks in an unexpected way. (that is, there is no check that guards against this, but it does eventually fail, usually)
```
// notice how it says struct instead of union
either := struct {
  a : int (<>);
  b : string (<>);
};
x := either(<>).a { 10 };
```

also, this one I think doesn't work, but it should (the fix for this is pretty big, so maybe it can be left for later).

```
a := b
c := a.id { 10 };
b := union { id : int(<>); };
```

SebastianMestre (approved):
Yeah, looks pretty good!

PR #200, opened 2020-11-23
Match expression

lushoBarlett:
Implements the match expression it looks something like this:
```rust
match(my_variant [: variant_mono]) {
    constructor_name1 { declaration [: inner_type] } => expression;
    constructor_name2 { declaration [: inner_type] } => expression;
    ...
};
```

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -183,6 +183,18 @@ struct TernaryExpression : public AST {
 	    : AST {ASTTag::TernaryExpression} {}
 };
 
+struct MatchExpression : public AST {
+	// NOTE: I'm avoiding using Declaration for this
+	// because it will generate confusion
+	Identifier m_matchee;
+	AST* m_type_hint {nullptr};
+	std::vector<Token const*> m_cases;
+	std::vector<AST*> m_expressions;
+
+	MatchExpression()
+	    : AST {ASTTag::MatchExpression} {}
+};
+
 struct ConstructorExpression : public AST {
 	AST* m_constructor;
 	std::vector<AST*> m_args;
--- a/src/compute_offsets.cpp
+++ b/src/compute_offsets.cpp
@@ -121,6 +121,13 @@ void compute_offsets(TypedAST::AccessExpression* ast, int frame_offset) {
 	compute_offsets(ast->m_record, frame_offset);
 }
 
+void compute_offsets(TypedAST::MatchExpression* ast, int frame_offset) {
+	compute_offsets(&ast->m_matchee, frame_offset);
+
+	for (auto& expr : ast->m_expressions)
+		compute_offsets(expr.second, frame_offset);
+}
+
 void compute_offsets(TypedAST::ConstructorExpression* ast, int frame_offset) {
 	compute_offsets(ast->m_constructor, frame_offset);
 
@@ -169,6 +176,7 @@ void compute_offsets(TypedAST::TypedAST* ast, int frame_offset) {
 		DISPATCH(CallExpression);
 		DISPATCH(TernaryExpression);
 		DISPATCH(AccessExpression);
+		DISPATCH(MatchExpression);
 		DISPATCH(ConstructorExpression);
 
 		DISPATCH(Block);
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -107,6 +107,19 @@ TypedAST::TypedAST* ct_eval(
 	return ast;
 }
 
+TypedAST::TypedAST* ct_eval(
+    TypedAST::MatchExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	// NOTE: no need to ct_eval the identifier, because it is guaranteed
+	// to be of metatype value, thus nothing needs to be done
+	if (ast->m_type_hint)
+		ast->m_type_hint = ct_eval(ast->m_type_hint, tc, alloc);
+
+	for (auto& expr : ast->m_expressions)
+		expr.second = ct_eval(expr.second, tc, alloc);
+
+	return ast;
+}
+
 TypedAST::ConstructorExpression* ct_eval(
     TypedAST::ConstructorExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
 	ast->m_constructor = constructor_from_ast(ast->m_constructor, tc, alloc);
@@ -357,6 +370,7 @@ TypedAST::TypedAST* ct_eval(
 		DISPATCH(IndexExpression);
 		DISPATCH(TernaryExpression);
 		DISPATCH(AccessExpression);
+		DISPATCH(MatchExpression);
 		DISPATCH(ConstructorExpression);
 
 		DISPATCH(Block);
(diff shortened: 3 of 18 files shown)


lushoBarlett:
Boy, did I write some code today! I'm quite happy about this. Turns out it was pretty trivial, like I suspected. It did require a handful of meddling with some data tho, but that was really fun.

Hope you don't mind the thing not working, I added the test and of course it just fails, but I did wanna add it as an example of what I want the test to be. Hopefully `compute_offsets` and `eval` are trivial as well (for you at least, I didn't wanna get into the offset thing).

src/typed_ast.hpp
```diff
@@ -226,6 +226,17 @@ struct TernaryExpression : public TypedAST {
 	    : TypedAST {TypedASTTag::TernaryExpression} {}
 };
 
+struct MatchExpression : public TypedAST {
+	// NOTE: I'm avoiding using Declaration for this
+	// because it will generate confusion
+	Identifier m_matchee;
+	TypedAST* m_type_hint {nullptr};
+	std::unordered_map<InternedString, FunctionLiteral*> m_expressions;
```
SebastianMestre, line 234: I don't think piggybacking on function literals is a good idea

src/ast.hpp
```diff
@@ -183,6 +183,18 @@ struct TernaryExpression : public AST {
 	    : AST {ASTTag::TernaryExpression} {}
 };
 
+struct MatchExpression : public AST {
+	// NOTE: I'm avoiding using Declaration for this
+	// because it will generate confusion
+	Identifier m_matchee;
```
SebastianMestre, line 189: We should be able to match on any expression, not only identifiers

src/metacheck.cpp
```diff
@@ -94,6 +94,22 @@ void metacheck(TypedAST::AccessExpression* ast, TypeChecker& tc) {
 		tc.m_core.m_meta_core.unify(ast->m_meta_type, tc.meta_value());
 }
 
+void metacheck(TypedAST::MatchExpression* ast, TypeChecker& tc) {
+	ast->m_meta_type = tc.meta_value();
```
SebastianMestre, line 129: This should probably be a call to the new metacheck assignment function

src/ct_eval.cpp
```diff
@@ -107,6 +107,19 @@ TypedAST::TypedAST* ct_eval(
 	return ast;
 }
 
+TypedAST::TypedAST* ct_eval(
+    TypedAST::MatchExpression* ast, TypeChecker& tc, TypedAST::Allocator& alloc) {
+	// NOTE: no need to ct_eval the identifier, because it is guaranteed
+	// to be of metatype value, thus nothing needs to be done
```
SebastianMestre, line 113: Since we should allow expressions in there (which can have type expressions within them), we should `ct_eval` the matchee

src/typecheck.cpp
```diff
@@ -207,6 +207,40 @@ void typecheck(TypedAST::AccessExpression* ast, TypeChecker& tc) {
...
+
+	std::unordered_map<InternedString, MonoId> dummy_structure;
+	for (auto& expr : ast->m_expressions) {
+		// unify return of match with return of functions
+		typecheck(expr.second, tc);
+		tc.m_core.m_mono_core.unify(ast->m_value_type, expr.second->m_return_type);
+
+		// get the structure of the match expression for a dummy
+		auto inner_value_declaration = expr.second->m_args.begin();
+		dummy_structure[expr.first] = inner_value_declaration->m_value_type;
```
SebastianMestre, line 227: I don't really understand what this is about...

src/typecheck.cpp
```diff
@@ -207,6 +207,40 @@ void typecheck(TypedAST::AccessExpression* ast, TypeChecker& tc) {
...
+
+	TypeFunctionId dummy_tf = tc.m_core.new_type_function(
+	    TypeFunctionTag::Sum,
+	    // we don't care about field order in dummies
+	    {},
+	    std::move(dummy_structure),
+	    true);
+
+	// NOTE: is it okay to assume 0 arguments?
+	// is f(a,b,c) === f'() possible?
```
SebastianMestre, line 238: We don't do polymorphic user defined types, so it's ok, for now. A comment mentioning this may be appropriate, though.

src/compute_offsets.cpp
```diff
@@ -121,6 +121,13 @@ void compute_offsets(TypedAST::AccessExpression* ast, int frame_offset) {
 	compute_offsets(ast->m_record, frame_offset);
 }
 
+void compute_offsets(TypedAST::MatchExpression* ast, int frame_offset) {
+	compute_offsets(&ast->m_matchee, frame_offset);
+
+	for (auto& expr : ast->m_expressions)
+		compute_offsets(expr.second, frame_offset);
+}
```
SebastianMestre, line 129: Since piggybacking on function literals is nasty, we should do something different here.

Something like starting a new block for every match-case, and 'pushing' space for the name that's used on every case.

SebastianMestre: You can think about what should happen by pretending to inline `compute_offsets` of function_literals and call_expressions into this function, and see what things become redundant.

example: we don't need captures because the call happens surrounded by the same scope as the literal creation.

lushoBarlett:
Right now there's two problems:
1. The type hints for the inner values seem to be ignored
2. We (wrongly) assume an identifier is used in match, which is obviously not true

src/interpreter/eval.cpp
```diff
@@ -276,6 +276,33 @@ void eval(TypedAST::AccessExpression* ast, Interpreter& e) {
...
+	e.m_env.start_stack_region();
+	e.m_env.push(variant_actually->m_inner_value);
+
+	e.m_env.start_stack_frame();
+	eval_call_function(static_cast<Function*>(function_val), 1, e);
+	e.m_env.end_stack_frame();
+
+	e.m_env.end_stack_region();
+
+	e.m_env.push(e.fetch_return_value());
```
SebastianMestre, line 303: Woah, it is pretty cool that the abstractions come together so nicely for this to be possible, but all the redundant work that goes on is kinda silly.

SebastianMestre (changes requested):
The way you implement match expressions in terms of functions is kinda cool, but I don't think it's good.

I would like it to be changed at some point, but I could be co0nvinced that it is good enough as a first implementation.

Also, I think matching on arbitrary expressions (and not only identifiers) should be a thing (I could also be convinced to delay this change for later, but a really good argument would be required)

Other than that, I left a few comments about things and answered some questions. cheers

src/compute_offsets.cpp
SebastianMestre: The same applies to `match_identifiers`

lushoBarlett:
Regarding the comments. I piggybacked on FunctionLiterals because it was easy. The problem is that if I change it, I kinda just throw 80% of the pr out the window. Should I just close it?

SebastianMestre:
> Right now there's two problems:
> 
>     1. The type hints for the inner values seem to be ignored

Well, yeah. we don't actually do anything with type hints in function arguments

>     2. We (wrongly) assume an identifier is used in match, which is obviously not true

I could use some clarfication here

lushoBarlett:
2 is the same as you said, match can work on things that are not identifiers. I knew that, I was gonna fix that. But well, the whole deal with using functions not being the right thing just made me close this.

Don't worry about this one we'll make another pr

PR #201, opened 2020-11-24
Naming cleanup

SebastianMestre:
We were using some inconsistent naming. That's gone now.

We always say 'record' to mean something with fields that can be accessed

We always say 'variant' to mean something that can be one of a set of types

We always say 'tuple' when talking about tuples (which aren't implemented)


lushoBarlett (approved):
Quality of life change

PR #202, opened 2020-11-24
Topologically ordered metacheck

SebastianMestre:
As I mentioned somewhere, 

```rust
a := c;
b := a.id { 10 };
c := union { id : int(<>); }(<>);
```

didn't work. This was due to not knowing the metatype of `a` when checking `a.id` in `c`, which made its metatype default to `value`, which is wrong (it should be `constructor`).

While I did not improve on that behavior, I made it harder to reach a declaration without knowing the metatypes of parts of that expression by metachecking declarations in topological order (aka in dependency order)

### DIFF
--- a/src/metacheck.cpp
+++ b/src/metacheck.cpp
@@ -162,22 +162,28 @@ void metacheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
 	for (auto& decl : ast->m_declarations)
 		decl.m_meta_type = tc.new_meta_var();
 
-	for (auto& decl : ast->m_declarations) {
-		if (decl.m_type_hint) {
-			metacheck(decl.m_type_hint, tc);
-			tc.m_core.m_meta_core.unify(decl.m_type_hint->m_meta_type, tc.meta_monotype());
+	auto const& comps = tc.m_env.declaration_components;
+	for (auto const& comp : comps) {
+
+		for (auto decl : comp) {
+			if (decl->m_type_hint) {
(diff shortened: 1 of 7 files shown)


SebastianMestre:
This is pretty cool: the changes in 9668466 make the changes in 36fd58e redundant.

This means 36fd58e could be reverted without loosing functionality. I don't think we should, though.

Doing things in topological order makes it so there is more (redundant) information when you process a declaration, which can be useful for generating error messages.

src/metacheck.cpp
```diff
@@ -162,22 +162,28 @@ void metacheck(TypedAST::DeclarationList* ast, TypeChecker& tc) {
...
-			tc.m_core.m_meta_core.unify(decl.m_type_hint->m_meta_type, tc.meta_monotype());
+	auto const& comps = tc.m_env.declaration_components;
+	for (auto const& comp : comps) {
+
+		for (auto decl : comp) {
+			if (decl->m_type_hint) {
+				metacheck(decl->m_type_hint, tc);
+				tc.m_core.m_meta_core.unify(
+				    decl->m_type_hint->m_meta_type, tc.meta_monotype());
+				// TODO? unify ( decl->metatype, value )
```
lushoBarlett, line 173: Yes.

lushoBarlett (approved):
All right, looks good. I like how you utilize the topological order 3 times, and the fact that it is stored in the environment.

To the fact that it isn't needed in `ct_eval`... Better to have it still.

PR #203, opened 2020-11-25
Parser cleanup

lushoBarlett:
Attempts to reduce many lines from the parser, without a change in speed. Basically a tiny refactor.

Maybe the littlest speed might get sacrificed in order to abstract some bits into auxiliary functions.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -8,11 +8,11 @@
 
 #include <cassert>
 
-#define HANDLE(result, writer)                                                 \
+#define HANDLE(writer)                                                         \
 	if (handle_error(result, writer))                                      \
 		return result;
 
-#define REQUIRE(result, token)                                                 \
+#define REQUIRE(token)                                                         \
 	if (handle_error(result, require(TokenTag::token)))                    \
 		return result;
 
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -76,12 +76,9 @@ Writer<AST::AST*> Parser::parse_top_level() {
 	    {"Parse Error: Failed to parse top level program"}};
 
 	auto declarations = parse_declaration_list(TokenTag::END);
-
-	if (handle_error(result, declarations))
-		return result;
+	HANDLE(declarations);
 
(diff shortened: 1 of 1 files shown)


lushoBarlett:
Further improvements will be added, mainly while loops and some common parse patterns. I want to merge #200 first because there's a new function there that I wanted to use here.

SebastianMestre:
Hey. I would recommend copying the function from the other branch, and dealing with possible conflicts at merge time.

lushoBarlett:
Fair point. I still have to decide what other changes I can do. Any ideas you have are welcome

src/parser.cpp
```diff
@@ -8,6 +8,14 @@
 
 #include <cassert>
 
+#define HANDLE(writer)                                                         \
+	if (handle_error(result, writer))                                      \
```
SebastianMestre, line 12: i dont like the name `HANDLE`, maybe something more descriptive, like `HANDLE_ERROR` or `CHECK_AND_RETURN`, like we use elsewhere, would be better.

src/parser.cpp
```diff
@@ -8,6 +8,14 @@
...
+		return result;
+
+#define REQUIRE(token)                                                         \
+	if (handle_error(result, require(TokenTag::token)))                    \
+		return result;
```
SebastianMestre, line 17: I think I prefer typing `TokenTag::` at the usage site. wdyt?

SebastianMestre: It also allows for things like

```c++
auto tag = TokenTag::something;
REQUIRE(tag);
```

SebastianMestre:
I'm not too sure about 6480327

lushoBarlett:
I was very unsure of some of my decisions here, so I will gladly accept your suggestions and apply them sometime soon

lushoBarlett:
I reverted the bad parts, and I'm gonna stop right here, because I have no clue what else to do. If it needs more cleanup, another pr can be done.

SebastianMestre (approved):
This looks much better to me

PR #209, opened 2020-12-19
Add file argument

lushoBarlett:
The text itself is TBD, I want your take, you may know standards a lot better than me.

### DIFF
--- a/src/interpreter/main.cpp
+++ b/src/interpreter/main.cpp
@@ -16,9 +16,18 @@
 #include "interpreter.hpp"
 #include "value.hpp"
 
-int main() {
+int main(int argc, char** argv) {
 
-	std::ifstream in_fs("source.jp");
+	if (argc < 2) {
+		std::cout << "Argument missing: source file" << std::endl;
+		return 0;
+	}
+
+	std::ifstream in_fs(argv[1]);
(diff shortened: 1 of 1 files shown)


src/interpreter/main.cpp
```diff
@@ -16,9 +16,18 @@
...
+int main(int argc, char** argv) {
 
-	std::ifstream in_fs("source.jp");
+	if (argc < 2) {
+		std::cout << "Argument missing: source file" << std::endl;
+		return 0;
+	}
+
+	std::ifstream in_fs(argv[1]);
+	if (!in_fs.good()) {
```
SebastianMestre, line 27: This will trigger if the file is empty, which you may or may not want to handle differently. (I'm ok with the way you did it, for the record)

The relevant documentation for the different failure modes of an `ifstream` can be seen here: https://en.cppreference.com/w/cpp/io/ios_base/iostate

SebastianMestre (approved):
Awesome. I'd like to have a nicer usage message eventually, but this is good for now.

something like

```
usage:
jasperi file [options]
```

or something. I don't really know

lushoBarlett:
Yea I was talking about the usage stuff primarily, we could do that in the future tho.

PR #210, opened 2020-12-20
Argument type hints

lushoBarlett:
Takes function arguments' type hints into account for type inference. As a sidenote, I consider that we may benefit on a way to test for program failures as well.

### DIFF
--- a/src/metacheck.cpp
+++ b/src/metacheck.cpp
@@ -48,8 +48,10 @@ void metacheck(TypedAST::FunctionLiteral* ast, TypeChecker& tc) {
 	assign_meta_type(ast->m_meta_type, tc.meta_value(), tc);
 
 	for (auto& arg : ast->m_args) {
-		if (arg.m_type_hint)
+		if (arg.m_type_hint) {
 			metacheck(arg.m_type_hint, tc);
+			assign_meta_type(arg.m_type_hint, tc.meta_monotype(), tc);
+		}
 		assign_meta_type(arg.m_meta_type, tc.meta_value(), tc);
 	}
 
--- a/src/metacheck.cpp
+++ b/src/metacheck.cpp
@@ -50,7 +50,7 @@ void metacheck(TypedAST::FunctionLiteral* ast, TypeChecker& tc) {
 	for (auto& arg : ast->m_args) {
 		if (arg.m_type_hint) {
 			metacheck(arg.m_type_hint, tc);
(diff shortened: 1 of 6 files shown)


SebastianMestre:
> ... we may benefit on a way to test for program failures as well.

That will be trivial once we start elevating type errors to the caller instead of asserting on them. For now, that's pretty much impossible (or at least really hard, by using multiprocessing and stuff).

SebastianMestre (changes requested):
This looks good to me, except for a tiny tiny slip in metacheck.

src/metacheck.cpp
```diff
@@ -47,10 +47,11 @@ void metacheck(TypedAST::Identifier* ast, TypeChecker& tc) {
 void metacheck(TypedAST::FunctionLiteral* ast, TypeChecker& tc) {
 	assign_meta_type(ast->m_meta_type, tc.meta_value(), tc);
 
-	int arg_count = ast->m_args.size();
-	for (int i = 0; i < arg_count; ++i)
-		// TODO: metacheck type hints
-		assign_meta_type(ast->m_args[i].m_meta_type, tc.meta_value(), tc);
+	for (auto& arg : ast->m_args) {
+		if (arg.m_type_hint)
+			metacheck(arg.m_type_hint, tc);
```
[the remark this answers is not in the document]
lushoBarlett: woops

src/metacheck.cpp
```diff
@@ -48,8 +48,10 @@ void metacheck(TypedAST::FunctionLiteral* ast, TypeChecker& tc) {
 	assign_meta_type(ast->m_meta_type, tc.meta_value(), tc);
 
 	for (auto& arg : ast->m_args) {
-		if (arg.m_type_hint)
+		if (arg.m_type_hint) {
 			metacheck(arg.m_type_hint, tc);
+			assign_meta_type(arg.m_type_hint, tc.meta_monotype(), tc);
```
SebastianMestre, line 53: should be

```c++
assign_meta_type(arg.m_type_hint->m_meta_type, tc.meta_monotype(), tc);
```

SebastianMestre (changes requested):

src/metacheck.cpp
lushoBarlett: This is what happens when I try and do it from github instead of turning on my other computer

SebastianMestre (approved):
Looks good to me

PR #211, opened 2021-01-01
Cleaning up

SebastianMestre:
Cleaning up some of the stuff that was mentioned in #208


SebastianMestre:
I'm happy with this PR as it is. It's not very focused, but I do think every diff is an improvement on something.

lushoBarlett (approved):
Looks good. Logging will be useful, being abstracted an all.

PR #212, opened 2021-01-01
Eval: store captures in a vector instead of a hash table

SebastianMestre:
It's been bugging me how we use a hash table to store captures in the runtime, so that's now gone.

This is certainly faster for programs that use lots of closures.


lushoBarlett (approved):
Nice simple speedup.

PR #215, opened 2021-01-03
Lexer improvement

SebastianMestre:
I moved the awful lexer stuff into a data structure.

It's horrendously inefficient in terms of memory usage, and probably a lot slower in runtime than before. I still think it's a win just due to the maintainability gains.

(edit)

You can think about this as an incomplete table based lexer. It might be a good idea to move to a full one eventually, but meh, baby steps.

### DIFF
--- a/src/algorithms/trie.cpp
+++ b/src/algorithms/trie.cpp
@@ -6,61 +6,61 @@ Trie::Node::Node() {
 	children.fill(nil);
 }
 
-Trie::Trie (){
+Trie::Trie() {
 	nodes.push_back(Node {});
 }
 
-uint16_t Trie::go(uint16_t n, uint8_t uc) {
-	return nodes[n].children[uc];
+uint8_t Trie::go(uint8_t node, uint8_t uc) {
+	return nodes[node].children[uc];
 }
 
 void Trie::insert(Entry entry) {
 	int entry_idx = entries.size();
 	entries.push_back(entry);
 
-	int n = 0;
+	int node = 0;
 	for (char c : entry.text) {
 		uint8_t uc = c;
-		auto nn = go(n, uc);
-		if (nn == nil) {
-			nn = nodes[n].children[uc] = nodes.size();
+		auto next_node = go(node, uc);
+		if (next_node == nil) {
+			next_node = nodes[node].children[uc] = nodes.size();
 			nodes.push_back(Node {});
 		}
 
-		n = nn;
+		node = next_node;
 	}
 
-	assert(nodes[n].entry == nil);
-	nodes[n].entry = entry_idx;
+	assert(nodes[node].entry == nil);
+	nodes[node].entry = entry_idx;
 }
 
 bool Trie::has(string_view str) {
-	int n = 0;
+	int node = 0;
 	for (char c : str) {
 		uint8_t uc = c;
-		auto nn = go(n, uc);
-		if (nn == nil)
+		auto next_node = go(node, uc);
+		if (next_node == nil)
 			return false;
 
-		n = nn;
+		node = next_node;
 	}
-	return nodes[n].entry != nil;
+	return nodes[node].entry != nil;
 }
 
 Trie::Entry Trie::longest_prefix_of(string_view haystack) {
-	uint16_t entry_idx = nil;
-	uint16_t n = 0;
+	uint8_t entry_idx = nil;
+	uint8_t node = 0;
 
 	for (char c : haystack) {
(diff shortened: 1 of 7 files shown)


src/lexer.hpp
```diff
@@ -52,4 +48,15 @@ struct Lexer {
...
+	std::vector<char> m_source;
+	TokenArray& m_tokens;
+
+	Trie m_symbols_trie;
+
+	int m_source_index {0};
+	int m_token_index {0};
+
+	int m_current_line {0};
+	int m_current_column {0};
```
lushoBarlett, line 61: Why did you move this down here?

src/lexer.cpp
```diff
@@ -102,244 +148,84 @@ std::pair<bool, TokenTag> Lexer::is_keyword(InternedString const& str) {
...
-		case '=':
-			push_token(TokenTag::DECLARE_ASSIGN, 2);
-			break;
-		default:
-			push_token(TokenTag::DECLARE, 1);
-		}
-		break;
+	char const* base_ptr = m_source.data() + i0;
+	m_tokens.push_back(
+			{TokenTag::STRING, InternedString {base_ptr, len}, l0, c0, m_current_line, m_current_column});
```
lushoBarlett, line 182: This seems indented weirdly.

src/algorithms/trie.hpp
```diff
@@ -0,0 +1,36 @@
...
+		int data;
+	};
+
+	struct Node {
+		Node();
+		Node(Node&&) = default;
+		Node(Node const&) = default;
+
+		uint16_t entry {nil};
+		std::array<uint16_t, 255> children;
```
lushoBarlett, line 22: Well this is... some memory usage.

src/algorithms/trie.cpp
```diff
@@ -0,0 +1,77 @@
+#include "trie.hpp"
+#include <iostream>
+#include <cassert>
+
+Trie::Node::Node() {
+	children.fill(nil);
+}
+
+Trie::Trie (){
```
lushoBarlett, line 9: Open bracket inconsistency

src/algorithms/trie.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void Trie::insert(Entry entry) {
+	int entry_idx = entries.size();
+	entries.push_back(entry);
+
+	int n = 0;
+	for (char c : entry.text) {
+		uint8_t uc = c;
+		auto nn = go(n, uc);
```
lushoBarlett, line 24: Perhaps more purpose indicative identifiers would be nice here.

lushoBarlett: And everywhere for that matter.

lushoBarlett (approved):
I looked at it a bunch, struggled a bit to understand, but ultimately found no bugs. I do agree to merge it, but I'm not happy with it as is.

The good thing about it at least is that for our purposes it's not super slow, because we use it for symbols and we don't use that many. The Lexer was never a bottleneck to begin with.

I do notice how you are leaning more towards abstraction instead of optimization now for some reason with the Lexer. I see it as a good balance to have, perhaps we did some very premature optimizations there that were not necessary yet.

src/lexer.hpp
lushoBarlett: I draw attention to this because it is inconsistent with the rest of the codebase.

SebastianMestre: I think we should move to that style everywhere.

When you look at a class definition, you more often want to know what the methods are. It's just more convenient to read.

src/lexer.cpp
SebastianMestre: Yeah I just cut and pasted without worrying about it

src/algorithms/trie.cpp
SebastianMestre: Lol it's very competitive programming-y

SebastianMestre: FWIW, c stands for char, uc for unsigned char, n for node, and nn for next node.

I'll change the names

SebastianMestre:
> I do notice how you are leaning more towards abstraction instead of optimization now for some reason with the Lexer. I see it as a good balance to have, perhaps we did some very premature optimizations there that were not necessary yet.

Meh, it was not optimized before, really. It was written in the most straightforward way possible, which happened to be *decently* fast (it was probably not branch prediction friendly, though).

I want to write a really fast lexer eventually, too, though.

PR #217, opened 2021-01-05
Bugfix capture match variable

SebastianMestre:
If we tried to capture the variable that gets bound in a match expression, eval would crash with an assertion stating that the captures were not references.

Now, we explicitly unbox and wrap the value in a fresh reference.

It might be better to not "re-box" it the value was already a reference, but that's a design decision that exceeds the scope of this pr

### DIFF
--- a/src/interpreter/eval.cpp
+++ b/src/interpreter/eval.cpp
@@ -293,7 +293,9 @@ void eval(TypedAST::MatchExpression* ast, Interpreter& e) {
 
 	// We won't pop it, because it is already lined up for the later
 	// expressions. Instead, replace the variant with its inner value.
-	e.m_env.m_stack.back() = variant_inner;
+	// We also wrap it in a reference so it can be captured
+	auto ref = e.new_reference(unboxed(variant_inner));
+	e.m_env.m_stack.back() = ref.get();
 	
 	auto case_it = ast->m_cases.find(constructor);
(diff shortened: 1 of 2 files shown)


lushoBarlett (approved):
Interesting bug. I'm curious how you discovered it.

I believe the way we hande values/references with the `unboxed` function is not very clean and is very prone to errors. Like this, and others in the past, we do tend to forget and I've always seen the code in eval as quite ugly. I'm interested in figuring out a way to handle that better sometime in the future. What do you think?

E: To clarify, I believe that the whole deal with unboxing and all that is a weak point of the codebase, and that should be addressed if we seek robustness.

SebastianMestre:
> Interesting bug. I'm curious how you discovered it.

I learned about "Continuation Passing Style" and tried using it in a jasper script. It crashed at runtime because of this bug

> I believe the way we hande values/references with the `unboxed` function is not very clean and is very prone to errors. Like this, and others in the past, we do tend to forget and I've always seen the code in eval as quite ugly. I'm interested in figuring out a way to handle that better sometime in the future. What do you think?

I would like values and references to be part of the type system, but I don't know how to approach that.

> E: To clarify, I believe that the whole deal with unboxing and all that is a weak point of the codebase, and that should be addressed if we seek robustness.

The whole of eval needs to be swapped out by a bytecode based interpreter imo

PR #218, opened 2021-01-05
Serialize ast

SebastianMestre:
I thought it might be cool for the AST printing to have a structured format, so I'm making it output S-expressions

I also added a new program that just dumps ASTs. I think if we extend it to do other things (reading the ast output format, desugaring, converting to typedAST, outputting readable code, etc. with flags to control it), it could be a valuable debugging and prototyping tool


lushoBarlett (approved):
Would be nice if we added the dispatch macro we use everywhere else. The rest is fine I guess. I don't really like the way you repeat `std::cout << "\n"` everywhere but whatever.

PR #221, opened 2021-01-07
Negative numbers

lushoBarlett:
Adds support for negative numbers, and as a bonus allows the redundant `+` sign as well. Not the best data storing approach, and some feature speculation was made.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -42,18 +42,18 @@ void print(Declaration* ast, int d) {
 
 void print(NumberLiteral* ast, int d) {
 	std::string stab(d - 1, tabc);
-	std::cout << stab << "[ Number ";
+	std::cout << stab << "(number-literal \"";
 	if (ast->m_negative)
 		std::cout << '-';
-	std::cout << ast->text() << " ]\n";
+	std::cout << ast->text() << "\")";
 }
 
 void print(IntegerLiteral* ast, int d) {
 	std::string stab(d - 1, tabc);
-	std::cout << stab << "[ Integer ";
+	std::cout << stab << "(integer-literal \"";
 	if (ast->m_negative)
 		std::cout << '-';
-	std::cout << ast->text() << " ]\n";
+	std::cout << ast->text() << "\")";
 }
 
 void print(StringLiteral* ast, int d) {
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -42,12 +42,16 @@ void print(Declaration* ast, int d) {
 
 void print(NumberLiteral* ast, int d) {
 	std::string stab(d - 1, tabc);
-	std::cout << stab << "[ Number " << ast->text() << " ]\n";
+	std::cout << stab << "[ Number "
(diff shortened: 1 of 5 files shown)


src/parser.cpp
```diff
@@ -428,6 +428,30 @@ Writer<AST::AST*> Parser::parse_terminal() {
...
+		} else if (match(TokenTag::NUMBER)) {
+			auto e = m_ast_allocator->make<AST::NumberLiteral>();
+			e->m_negative = token->m_type == TokenTag::SUB;
+			e->m_sign = token;
+			e->m_token = peek();
+			m_lexer->advance();
+			return make_writer<AST::AST*>(e);
+		}
+
+		return result;
```
SebastianMestre, line 452: It would be nice to push a specific error message, indicating that there was a stray prefix '+' or ' -' with no number

src/ast.hpp
```diff
@@ -44,6 +44,8 @@ struct DeclarationList : public AST {
 };
 
 struct IntegerLiteral : public AST {
+	bool m_negative {false};
+	Token const* m_sign {nullptr};
```
SebastianMestre, line 48: Ahh that's janky but whatever. It's not like keeping `AST::AST` tidy matters

src/ast.cpp
```diff
@@ -42,12 +42,18 @@ void print(Declaration* ast, int d) {
...
+	std::cout << ast->text() << " ]\n";
 }
 
 void print(IntegerLiteral* ast, int d) {
 	std::string stab(d - 1, tabc);
-	std::cout << stab << "[ Integer " << ast->text() << " ]\n";
+	std::cout << stab << "[ Integer ";
+	if (ast->m_negative)
+		std::cout << '-';
+	std::cout << ast->text() << " ]\n";
```
SebastianMestre, line 56: I know it's silly, but can you write this in the style of #218 ?

src/test/main.cpp
```diff
@@ -705,6 +705,44 @@ void interpreter_tests(Test::Tester& tests) {
...
+		    +[](Interpreter::Interpreter& env) -> ExitStatusTag {
+			    return Assert::equals(eval_expression("f", env), -1.1);
+		    },
+		    +[](Interpreter::Interpreter& env) -> ExitStatusTag {
+			    return Assert::equals(eval_expression("g", env), 1);
+		    },
+		    +[](Interpreter::Interpreter& env) -> ExitStatusTag {
+			    return Assert::equals(eval_expression("h", env), 1.1);
+		    }
+	    }));
```
SebastianMestre, line 745: Nice test. You earn 10 good boy points.

SebastianMestre (approved):
Hey that's cool.

For the future: maybe we should have a more general prefix operator `-`?

lushoBarlett:
> maybe we should have a more general prefix operator `-`?

Yea, this was the easy approach

src/ast.cpp
lushoBarlett: I don't really see the point? but okay

PR #222, opened 2021-01-08
Eval cleanup

SebastianMestre:
Addresses some of the points we discussed today


lushoBarlett (approved):
So much prettier! I thought you weren't gonna do this. I do have to say that this brings joy.

PR #223, opened 2021-01-09
Lexer: change the poly paren tokens

SebastianMestre:
I changed `(<` and `>)` to `<:` and `:>`, respectively.

I am still not happy, but I like it a lot more than before.


lushoBarlett (approved):
Empty poly parens are really weird looking, I do admit. Perhaps some day `<>` will be a reality?

PR #224, opened 2021-01-09
Eval: implement assignment to an array element

SebastianMestre:
We now put every array element in a reference. This is not the nicest
in terms of performance (because it adds an extra level of indirection),
but that's ok.

Also, I added a test.


lushoBarlett (approved):
Seems fair for a first implementation.

PR #227, opened 2021-01-10
Interpreter cleanup

SebastianMestre:
Continuing with the theme of #222, I'm doing some more cleaning up all over the interpreter.


SebastianMestre:
I'm about to get real busy, so let's get this over with.

lushoBarlett (approved):
I like these new abstractions and name changes. Looks cleaner to me.

PR #230, opened 2021-01-14
Bytecode interpreter

SebastianMestre:
We talked about this. A bytecode interpreter should be faster and stuff. (edit, disclaimer: it probably won't be much faster (or at all) at first, since we have inefficiencies in places other than just running the code)

Very WIP, but It already does some stuff, which is neat.


SebastianMestre:
my wrists hurt, presumably due to too much coding, so I'm stopping here for today

SebastianMestre:
Refactoring the tree interpreter, and gaining a better understanding of the order it does operations in, really helped me implement function calls lol

PR #231, opened 2021-01-14
Test rework

lushoBarlett:
Moves the tests to files to easily modify, and other simplifications.

PR #233, opened 2021-01-16
Function call simplification

### LINKED ISSUE #232: Bad stack handling in seq expressions
```rust
__invoke := fn() {
	y := 5 + seq { a := 1; b := 1; return b; };
	return y;
};
```
This program returns 10, but it should return 6.

SebastianMestre:
I changed the calling conventions: now arguments come after the frame pointer, followed by captures, and the callee does not get popped off the stack before the call, but right at the end. I also simplified a very convoluted part of the function call sequence, eliminating a hash table access per argument passed to a function.

With these changes, I observed an approximately 2x speed up in a call-heavy benchmark (computing Ackermann(3,3)).

Furthermore, I sneaked in a fix for #232. (closes #232)

There are a few more misc. changes sprinkled in, but those are not really worth mentioning.


src/interpreter/stack.cpp
```diff
@@ -57,6 +61,7 @@ Value*& Stack::access(int offset) {
 }
 
 Value*& Stack::frame_at(int offset) {
+	assert(m_frame_ptr + offset < m_stack.size());
 	return m_stack[m_frame_ptr + offset];
 }
```
lushoBarlett, line 70: As offsets can be negative, a check for lower bound would also be nice.

src/interpreter/eval.cpp
```diff
@@ -153,13 +153,8 @@ void eval_call_function(gc_ptr<Function> callee, int arg_count, Interpreter& e)
...
 	}
 
-	// this feels really dumb:
-	// we get a value on the stack, then we move it to the return value
-	// slot, then we pop some stuff off the stack, and put it back on the
-	// stack. It is doubly dumb when we eval a seq-expr (because it does a
-	// save-pop sequence)
 	eval(callee->m_def->m_body, e);
-	e.save_return_value(e.m_stack.pop_unsafe());
+	e.m_stack.frame_at(-1) = e.m_stack.pop_unsafe();
```
lushoBarlett, line 148: The `frame_at(-1)` could potentially be the new "store return value" and be called with such nomenclature. Shame that we still have something with that name and that we will change all this to bytecode anyway. So don't pay attention to this comment.

src/compute_offsets.cpp
```diff
@@ -43,9 +43,9 @@ void compute_offsets(TypedAST::IfElseStatement* ast, int frame_offset) {
 }
 
 void compute_offsets(TypedAST::CallExpression* ast, int frame_offset) {
-	compute_offsets(ast->m_callee, frame_offset);
+	compute_offsets(ast->m_callee, frame_offset++);
 	for (auto& arg : ast->m_args)
-		compute_offsets(arg, frame_offset);
+		compute_offsets(arg, frame_offset++);
 }
```
lushoBarlett, line 49: No need to presume a fix, write a test for it. The one in #232 specifically. You could merge #231, if you haven't already, and it would be super easy to do so. Add it to `tests/function.jp`.

lushoBarlett (approved):
Looks good, my only real complaint is the lack of a test for the bug fix. I will approve anyway to speed up the process of merge.

src/compute_offsets.cpp
SebastianMestre: Oh, I tested it. I said presumably because there might be other bugs of the same nature roaming.

PR #234, opened 2021-01-17
Log streams

SebastianMestre:
I did some more work on the logging lib. In particular, I added stream versions of all the loggers, and added support for InternedString.


lushoBarlett (approved):
Looks like an improvement. Not sure if the logging `cpp` files actually need to exist. Perhaps it's better to compress it all into the headers?

SebastianMestre:
> Perhaps it's better to compress it all into the headers?
It's not.

It's not. The way it is, consumers of `log.hpp` don't need to include `<iostream>`

lushoBarlett:
> It's not. The way it is, consumers of `log.hpp` don't include `<iostream>`

Yea I realized that later hahaha

PR #235, opened 2021-01-18
Bugfix NDEBUG lexer

SebastianMestre:
when compiling with `-DNDEBUG` (which eliminates assertions from source code, for faster performance on release builds), the lexer hung, due to not detecting comments and strings.

EDIT. I could not contain myself, and ended up optimizing `StringSet`. It must be about 2x faster now, because the tests take about 4% less time to run.


lushoBarlett (approved):
Wow, totally slipped under the radar. Nice replacement for modulo operator when the modulo is a power of 2, will remember that.

SebastianMestre:
> Wow, totally slipped under the radar.

I was being cute when I wrote that. It didn't even cross my mind that it would be a problem if we ever wanted to compile with `-DNDEBUG` at the time.

> Nice replacement for modulo operator when the modulo is a power of 2, will remember that.

Ah yeah, its an nifty performance boost. I should add a comment about it, though.

PR #236, opened 2021-01-19
Parser: don't add implicit poly parens in type terms

SebastianMestre:
This is very overdue lol. The fix was extremely simple, the parser is now one line shorter.

(EDIT) The reason this is a problem is that it forbids things like

```rust
Int := int<::>;
f := fn(x : Int) => x;
```


lushoBarlett (approved):
Finally!

Issue #238, opened 2021-01-20
Parentheses around if expressions are ugly


SebastianMestre:
I don't like the parentheses around if-expressions. It would be neat if we could get rid of them.

Example from the readme:

```rust
or_default := fn(value, default) =>
	if value == ""
		then default
		else value;
```

lushoBarlett:
I'm on it. When I figure out a nice way to check for disambiguation it will be done.

PR #239, opened 2021-01-21
Remove parenthesis from ternary

lushoBarlett:
Solves #238. Tangentially adds all expressions as statements in the most straight forward way.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -285,14 +285,17 @@ Writer<std::vector<AST::AST*>> Parser::parse_argument_list() {
  * Here is an article with more information:
  * https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
  */
-Writer<AST::AST*> Parser::parse_expression(int bp) {
+Writer<AST::AST*> Parser::parse_expression(int bp, AST::AST* parsed_lhs) {
 	Writer<AST::AST*> result = {
 	    {"Parse Error: Failed to parse expression"}};
 
 	Writer<AST::AST*> lhs;
-
-	lhs = parse_terminal();
-	CHECK_AND_RETURN(result, lhs);
+	if (!parsed_lhs) {
+		lhs = parse_terminal();
+		CHECK_AND_RETURN(result, lhs);
+	} else {
+		lhs = make_writer(parsed_lhs);
+	}
 
 	while (1) {
 		auto op = peek();
@@ -547,14 +550,20 @@ Writer<AST::AST*> Parser::parse_terminal() {
 	return result;
 }
 
-Writer<AST::AST*> Parser::parse_ternary_expression() {
+Writer<AST::AST*> Parser::parse_ternary_expression(AST::AST* parsed_condition) {
 	Writer<AST::AST*> result = {
 	    {"Parse Error: Failed to parse ternary expression"}};
 
-	REQUIRE(result, TokenTag::KEYWORD_IF);
+	Writer<AST::AST*> condition;
+	if (!parsed_condition) {
+		REQUIRE(result, TokenTag::KEYWORD_IF);
 
-	auto condition = parse_expression();
-	CHECK_AND_RETURN(result, condition);
+		condition = parse_expression();
+		CHECK_AND_RETURN(result, condition);
+	} else {
+		condition = make_writer(parsed_condition);
+	}
+
 	REQUIRE(result, TokenTag::KEYWORD_THEN);
 
 	auto then_expr = parse_expression();
@@ -756,16 +765,21 @@ Writer<AST::AST*> Parser::parse_return_statement() {
 	return make_writer<AST::AST*>(e);
 }
 
-Writer<AST::AST*> Parser::parse_if_else_statement() {
+Writer<AST::AST*> Parser::parse_if_else_statement(AST::AST* parsed_condition) {
 	Writer<AST::AST*> result = {
 	    {"Parse Error: Failed to parse if-else statement"}};
 
-	REQUIRE(result, TokenTag::KEYWORD_IF);
-	REQUIRE(result, TokenTag::PAREN_OPEN);
+	Writer<AST::AST*> condition;
+	if (!parsed_condition) {
+		REQUIRE(result, TokenTag::KEYWORD_IF);
+		REQUIRE(result, TokenTag::PAREN_OPEN);
 
-	auto condition = parse_expression();
-	CHECK_AND_RETURN(result, condition);
-	REQUIRE(result, TokenTag::PAREN_CLOSE);
+		auto condition = parse_expression();
+		CHECK_AND_RETURN(result, condition);
+		REQUIRE(result, TokenTag::PAREN_CLOSE);
+	} else {
+		condition = make_writer(parsed_condition);
+	}
 
 	auto body = parse_statement();
 	CHECK_AND_RETURN(result, body);
@@ -946,6 +960,7 @@ Writer<AST::AST*> Parser::parse_statement() {
 		CHECK_AND_RETURN(result, return_statement);
 		return return_statement;
 	} else if (p0->m_type == TokenTag::KEYWORD_IF) {
+		// first disambiguation: no parenthesis
 		auto* p1 = peek(1);
 		if (p1->m_type != TokenTag::PAREN_OPEN) {
 			auto expression = parse_expression();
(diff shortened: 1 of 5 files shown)


SebastianMestre:
I think there might be a better way to do this disambiguation: why not add an `AST::ParenthesizedExpression` node, which we return when there is an expression inside parens, then parsing if-else goes as follows (pseudocode):

```c++
auto parse_if_else(){
  Writer<AST*> result;
  REQUIRE(TokenTag::IF);
  expr = parse_expression();
  if (consume(TokenTag::Then)) {
    // continue parsing if-else expr
  } else if (expr->type() == ASTTag::ParenthesizedExpression) {
    // continue parsing if-else stmt
  } else {
    // failed to parse, push an error
    return result;
  }
}
```

What you did looks good, tho. I just think this would be a better way to factor it.

lushoBarlett:
So, having both ternary and if else statement be in the same parsing function?

src/parser.cpp
```diff
@@ -488,12 +491,9 @@ Writer<AST::AST*> Parser::parse_terminal() {
 		return function;
 	}
 
-	if (token->m_type == TokenTag::PAREN_OPEN and
-	    peek(1)->m_type == TokenTag::KEYWORD_IF) {
-		m_lexer->advance();
+	if (token->m_type == TokenTag::KEYWORD_IF) {
 		auto ternary = parse_ternary_expression();
 		CHECK_AND_RETURN(result, ternary);
-		REQUIRE(result, TokenTag::PAREN_CLOSE);
```
SebastianMestre, line 496: I love how this is shorter and clearer

SebastianMestre:
if else expressions would still need their function, because they can show within expressions, but yeah. This way, we don't need a function just for if else statements

src/parser.cpp
```diff
@@ -952,7 +960,39 @@ Writer<AST::AST*> Parser::parse_statement() {
...
 		return return_statement;
 	} else if (p0->m_type == TokenTag::KEYWORD_IF) {
-		auto if_else_statement = parse_if_else_statement();
+		// first disambiguation: no parenthesis
+		auto* p1 = peek(1);
+		if (p1->m_type != TokenTag::PAREN_OPEN) {
+			auto expression = parse_expression();
+			CHECK_AND_RETURN(result, expression);
+			REQUIRE(result, TokenTag::SEMICOLON);
+			return expression;
```
SebastianMestre, line 969: if it's not PAREN_OPEN, should it not just call `parse_ternary_expression`?

src/parser.cpp
```diff
@@ -952,7 +960,39 @@ Writer<AST::AST*> Parser::parse_statement() {
...
+		}
+
+		m_lexer->advance();
+
+		// we must include parenthesis otherwise
+		REQUIRE(result, TokenTag::PAREN_OPEN);
```
SebastianMestre, line 975: NIT: parenthesis is singular, parentheses is plural

lushoBarlett:
I wanted to avoid things like this because of code duplication, but I guess your way is quite good. My approach is not preferred, in mi opinion.

At least it works, and it a relatively speedy fashion.

src/parser.cpp
SebastianMestre: Sorry, it should not. my mistake

src/parser.cpp
```diff
@@ -285,14 +285,17 @@ Writer<std::vector<AST::AST*>> Parser::parse_argument_list() {
...
 	Writer<AST::AST*> lhs;
-
-	lhs = parse_terminal();
-	CHECK_AND_RETURN(result, lhs);
+	if (!parsed_lhs) {
+		lhs = parse_terminal();
+		CHECK_AND_RETURN(result, lhs);
+	} else {
+		lhs = make_writer(parsed_lhs);
+	}
```
SebastianMestre, line 298: I didn't quite understand why this was needed at first, but I do now. That's some good thinking!

SebastianMestre (commented):
Yeah, it looks pretty good. Let me know if you intend on using the factoring I mentioned, or you're keeping this one.

If you're keeping this one, I'll approve it.

src/parser.cpp
```diff
@@ -765,37 +767,56 @@ Writer<AST::AST*> Parser::parse_return_statement() {
...
-		auto condition = parse_expression();
-		CHECK_AND_RETURN(result, condition);
+	REQUIRE(result, TokenTag::KEYWORD_IF);
+	if (consume(TokenTag::PAREN_OPEN)) {
+		auto expr = parse_expression();
+		CHECK_AND_RETURN(result, expr);
 		REQUIRE(result, TokenTag::PAREN_CLOSE);
+
+		auto paren = m_ast_allocator->make<AST::ParenthesizedExpression>();
+		paren->m_expr = expr.m_result;
```
SebastianMestre: this looks exactly like inlining a piece of parse_terminal, why is calling parse_expression instead of writing this if-else here not good?

SebastianMestre: I mean, (indirectly) almost the first thing parse_expression does is check for paren and start building a paren-expr

SebastianMestre: also, does your code handle `if (func_name)(arg1, arg2) then value1 else value2` ?

SebastianMestre (commented):
I like this, but I don't feel comfortable accepting it.

SebastianMestre:
here is a diff over commit 806d3dc that I think does the right thing

```diff
diff --git a/src/parser.cpp b/src/parser.cpp
index 1542c04..f1dbe2b 100644
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -307,7 +307,9 @@ Writer<AST::AST*> Parser::parse_expression(int bp, AST::AST* parsed_lhs) {
                    op->m_type == TokenTag::PAREN_CLOSE ||
                    op->m_type == TokenTag::COMMA ||
                    op->m_type == TokenTag::KEYWORD_THEN ||
-                   op->m_type == TokenTag::KEYWORD_ELSE) {
+                   op->m_type == TokenTag::KEYWORD_ELSE ||
+                       op->m_type == TokenTag::KEYWORD_IF ||
+                       op->m_type == TokenTag::KEYWORD_RETURN) {
                        break;
                }
 
@@ -771,31 +773,16 @@ Writer<AST::AST*> Parser::parse_if_else_stmt_or_expr() {
        Writer<AST::AST*> result = {
            {"Parse Error: Failed to parse if-else statement"}};
 
-       AST::AST* condition;
-
        REQUIRE(result, TokenTag::KEYWORD_IF);
-       if (consume(TokenTag::PAREN_OPEN)) {
-               auto expr = parse_expression();
-               CHECK_AND_RETURN(result, expr);
-               REQUIRE(result, TokenTag::PAREN_CLOSE);
-
-               auto paren = m_ast_allocator->make<AST::ParenthesizedExpression>();
-               paren->m_expr = expr.m_result;
-
-               condition = paren;
-       } else {
-               auto expr = parse_expression();
-               CHECK_AND_RETURN(result, expr);
-
-               condition = expr.m_result;
-       }
+       auto condition = parse_expression();
+       CHECK_AND_RETURN(result, condition);
 
-       if (condition->type() == ASTTag::ParenthesizedExpression && !match(TokenTag::KEYWORD_THEN)) {
+       if (condition.m_result->type() == ASTTag::ParenthesizedExpression && !match(TokenTag::KEYWORD_THEN)) {
                auto body = parse_statement();
                CHECK_AND_RETURN(result, body);
 
                auto e = m_ast_allocator->make<AST::IfElseStatement>();
-               e->m_condition = condition;
+               e->m_condition = condition.m_result;
                e->m_body = body.m_result;
 
                if (consume(TokenTag::KEYWORD_ELSE)) {
@@ -809,7 +796,7 @@ Writer<AST::AST*> Parser::parse_if_else_stmt_or_expr() {
        }
 
        // rollback to whole expression
-       auto ternary = parse_ternary_expression(condition);
+       auto ternary = parse_ternary_expression(condition.m_result);
        CHECK_AND_RETURN(result, ternary);
 
        auto expression = parse_expression(0, ternary.m_result);
```

src/parser.cpp
SebastianMestre: I just tested it, it does not.

lushoBarlett:
Does force pushing a branch deactivate continuous integration? Why is it not checking my build?

SebastianMestre:
CircleCI is down atm

SebastianMestre (approved):

SebastianMestre:
no comment, it's perfect

PR #242, opened 2021-01-21
Bugfix global seq exprs

### LINKED ISSUE #240: Nested seq expressions are broken
Exhibit 1:
```rust
a := seq {
    b := seq { return 0; };
    return b;
};
```
This will tell me `Assertion 'e.m_return_value' failed`. I don't know which one fails.

Exhibit 2:
```rust
a := seq {
    var := 0;
    b := seq {
        var = 1;
        return 0;
    };
    return b;
};
```
This fails while casting in `as<T>`. It tells me that `Assertion 'x' failed.`, meaning the pointer itself is `nullptr`. I have tried other kinds of sequence expression capture tests, more normal ones, and they work fine.

Further problems have not yet been found, I suppose that's it.

SebastianMestre:
closes #240

We were deciding that some variables were global, when in reality they were local to a seq-expr

### DIFF
--- a/src/test/main.cpp
+++ b/src/test/main.cpp
@@ -128,15 +128,15 @@ void interpreter_tests(Test::Tester& tests) {
 	        EQUALS("__invoke()", 42)
 	    }));
 
-	tests.add_test(std::make_unique<Test::InterpreterTestSet>("tests/seq_expressions.jp",
+	tests.add_test(std::make_unique<Test::InterpreterTestSet>(
+	    "tests/seq_expressions.jp",
 	    Testers {
 	        EQUALS("return_const", 31415),
 	        EQUALS("return_call", 42),
-			EQUALS("issue232_1()", 6),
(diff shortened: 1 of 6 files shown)


src/test/main.cpp
```diff
@@ -130,6 +127,16 @@ void interpreter_tests(Test::Tester& tests) {
 	    Testers {
 	        EQUALS("__invoke()", 42)
 	    }));
+
+	tests.add_test(std::make_unique<Test::InterpreterTestSet>("tests/seq_expressions.jp",
+	    Testers {
+	        EQUALS("return_const", 31415),
+	        EQUALS("return_call", 42),
+			EQUALS("issue232_1()", 6),
+			EQUALS("issue232_2()", 7),
```
lushoBarlett, line 136: bad indent

lushoBarlett (approved):
Good fix.

PR #243, opened 2021-01-21
change order of operations in typefunc unification

### LINKED ISSUE #241: jasperi segfaults when running certain code
The code is attached below: it works correctly as is, but crashes with a segfault when the indicated type hint is removed. The segfault is caused by too much recursion in the type checker.

```rust
expr := union {
	sum : two_exprs;
	val : int<::>;
}<::>;

two_exprs := struct {
	left : expr;
	right : expr;
}<::>;

// removing the type hint makes it segfault
eval := fn(e : expr) => match(e) {
	val { v } => v;
	sum { v } => eval(v.left) + eval(v.right);
};

__invoke := fn() =>
	expr.sum { two_exprs {
		expr.val{ 10 };
		expr.val{ 32 };
	} } |> eval();
```

SebastianMestre:
closes #241 

This was quite the brain twister. The fix was surprisingly simple, though.

When unifying typefuncs, we would first go through unifying the members, and then convert one of the typefuncs into a var that points at the other. However, if there were cycles between the typefuncs, we would end up recursing down that cycle for ever.

We now, first, convert one of the typefuncs into a var, and then unify the members. Once we recurse down the cycle, we find that we are trying to unify a typefunc with a var to itself, and stop successfully

This change amounts to simply reordering two bits of code.

### DIFF
--- a/src/typesystem.cpp
+++ b/src/typesystem.cpp
@@ -6,6 +6,7 @@
 
 TypeSystemCore::TypeSystemCore() {
 	m_tf_core.unify_function = [this](Unification::Core& core, int a, int b) {
+		// TODO: Redo this once we add polymorphic records
 		if (core.find_term(a) == core.find_term(b))
 			return;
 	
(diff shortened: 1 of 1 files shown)


lushoBarlett (approved):
That's actually quite sensible. It's of the same nature as the vars we attach to declarations before checking their values.

PR #244, opened 2021-01-24
AST, TypedAST -> CST, AST naming

SebastianMestre:
Dude, the name 'TypedAST' is disgusting. I've been wanting to see it go since I first wrote it down. Now, I finally know what names to use instead: CST, for 'Concrete Syntax Tree', and AST, for 'Abstract Syntax Tree'.

Also, since it's abstract, the AST should not know jack about tokens and stuff. This also saves an indirection in many places.

> Side note: github doesn't understand successive renames where one file has the old name of another, so the diff ends up larger than the real change. You should probably look through individual commits instead of looking at the full patch.
>
> The first two commits are essentially find&replace anyways


lushoBarlett (approved):
Nice naming

PR #249, opened 2021-01-27
merge desugar and convert_ast

### LINKED ISSUE #248: desugar and convert_ast feel redundant
I feel like these two steps (`desugar` and `convert_ast`) could be merged.

Going forward, I want to add more information about its source code to every CST node. This means that if `desugar` returns a CST node, we would have to either leave that information blank, or make the values up.

Also, to give good error messages, we would have to store pointers to the non-desugared node on CSTs. Which means we would need to do several (at least 2, but it could be more) pointer jumps to get from an AST to its source information.

Instead, if we just do desugaring and conversion in a single step, we won't have these issues.

To be sure if this made sense, I asked online. It looks like almost everyone does it this way.

SebastianMestre:
closes #248 

I was sneaky and also removed dictionaries


lushoBarlett (approved):
Oh yes, that's a lot of lines you removed there. I'm very happy about that. Also removing dictionaries is something that I wanted.

SebastianMestre:
> Oh yes, that's a lot of lines you removed there. I'm very happy about that. Also removing dictionaries is something that I wanted.

uwu

PR #250, opened 2021-02-09
Document the high level achitecture of the project

SebastianMestre:
I added an architecture.md file with a high level description of the project. Chip in if there is anything else that should be in there.

The idea is to help potential new contributors. It's intentionally vague, so it doesn't require much maintenance.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,57 @@
+# Programs
+
+Within this repository, there are 3 different programs: the interpreter, the
+test suite, and the playground. Respectively, these 3
+
+- run jasper programs
+- run tests, which can be found under the `tests/` directory
+- show cst dumps and other useful debugging features
+
+# Stages
+
+The general stages of running a Jasper program go as follows:
+
+ - Parsing -- The parser produces a `CST`. This stores information about the
+ source location of the different syntactic constructs in a program.
+ - AST Conversion -- We convert the `CST` into an `AST`, discarding source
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -0,0 +1,57 @@
...
+the other interesting files are within the `src/` directory.
+
+The tree-walk interpreter lives in the `src/interpreter/` directory.
+
+There are a few generally useful data structures in the `src/utils/` directory,
+and some more project-specific ones in the `src/algorithms/` directory.
+
+`src/test/` stores the implementation of the test runner, while `src/tests/`
+stores the content of the tests themselves. You will need to touch these in
+order to add new tests.
```
lushoBarlett, line 57: `tests/` stores the tests themselves

lushoBarlett (commented):
I feel like this lacks our data model. I don't know how non-specific you want to get, but I don't think we are going to change our current data layout very soon anyway.

What I mean is, the fact that we have a `CST`, and we deal with polymorphism using CST tags, and the same with `AST`, and the same with `Value`. Mentioning the fact that we deal with those 3 kinds of structures and that we have tags for them, because I feel that is not very common among C++ projects.

This of course is up to you to decide. I could go either way

ARCHITECTURE.md
SebastianMestre: fixed

lushoBarlett (approved):

PR #251, opened 2021-02-15
First steps towards allowing dynamic typing

SebastianMestre:
We now use a bit less TC information at runtime. I believe this can be brought down to 0, easily enough. As of now, all our tests run to completion without the need to run the typechecker.

I also did a few more changes to make this and some future changes easier.

- Store strings instead of identifiers in AST::UnionExpression and AST::StructExpression
- Run eval in dependency order
- Evaluate the runtime representation of record and variant types from the non `ct_eval`'d AST nodes

Not yet done, but should be relatively easy to do:

- Remove the fields vector from `TypeFunctionData`, replace with a pointer to `AST`?
- Rethink/rework the `AST::Constructor` abstraction. Though originally envisioned to represent a variant constructor, it is also used for records.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -330,8 +330,8 @@ AST* convert_ast(CST::UnionExpression* cst, Allocator& alloc) {
 	auto ast = alloc.make<UnionExpression>();
 
 	for (auto& constructor : cst->m_constructors) {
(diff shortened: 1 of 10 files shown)


lushoBarlett (approved):
You've shown this to me before. It looks good

PR #252, opened 2021-02-18
Eliminate parse header

SebastianMestre:
I shuffled things around a bit. I put the `parse.hpp` declarations in `parser.hpp`, which allowed moving the `::Parser` class to `parser.cpp`, hiding it from the outer world. Then, since the only reason a `::Parser` stored a `::Lexer` by pointer, was to reduce transitive dependencies on `lexer.hpp`, and that problem no longer exists, we now put the `Lexer` directly in the `Parser`.

I touched up some internals of the lexer to do tokenizing up front, instead of doing it on-demand, as the parser needs tokens. I imagine this is slightly faster, and, more importantly, I'm sure it will let us write the lexer, as well as the parser's API more neatly in the future.

The current API:
```c++
Cst* parse(std::string const&, TokenArray&, Cst::Allocator&);

// usage:

std::string source = read_file();
TokenArray tokens;
Cst::Allocator allocator;
auto cst = parse(source, tokens, allocator);
```

A possibliity for future API design:
```c++
TokenArrray tokenize(std::string const&);
Cst* parse(TokenArray const&, Cst::Allocator&);

// usage:

std::string source = read_file();
auto tokens = tokenize(source);
Cst::Allocator allocator;
auto cst = parse(tokens, allocator);
```

Of course, the two are equivalent, but the later is made up of simpler pieces, more value-oriented and more in line with Modern C++ style.


lushoBarlett (approved):
A lot cleaner and modularized than before. Always nice.

PR #253, opened 2021-02-28
desugar for statements into while statements

SebastianMestre:
We now have a bit more desugaring code, but we have one fewer node type to handle on every later stage.

For loops might be a tiny little bit slower now, due to more indirections, but that will go away once we compile to bytecode.


lushoBarlett (approved):
Makes sense. Maybe there's other desugaring that can be done, for bytecode purposes the less the better.

PR #254, opened 2021-02-28
remove unneeded references in records and variants

SebastianMestre:
I was just writing some code, and found we missed some reference unwrappings in eval. It's probably not too hard to write some code that does something unexpected using this, but I don't think that's too interesting, or needed.


lushoBarlett (approved):
Always better to have a `value_of`

PR #255, opened 2021-03-06
Symbol table

SebastianMestre:
`CompileTimeEnvironment` was handling too many things, so I saw an opportunity for separation: it wasn't that hard to tell that most of its methods were only used on `match_identifiers`.

Once It was separated, I changed the data structure, to make it faster. One of the insights is that only nested scopes were ever used, so it could be optimized around that fact.

name lookups are now a single hash table lookup, and not some messy, branchy, loop over a vector of hash tables.

### DIFF
--- a/src/symbol_table.cpp
+++ b/src/symbol_table.cpp
@@ -17,8 +17,7 @@ SymbolTable::SymbolMap& SymbolTable::latest_shadowed_scope() {
 
 void SymbolTable::declare(AST::Declaration* decl) {
 	auto old_decl = [&]() -> AST::Declaration* {
-		auto insert_result =
-		    m_available_vars.insert({decl->identifier_text(), decl});
+		auto insert_result = m_bindings.insert({decl->identifier_text(), decl});
 		if (insert_result.second) // introduced a previously undeclared name
 			return nullptr;
 		return std::exchange(insert_result.first->second, decl);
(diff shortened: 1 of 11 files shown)


src/symbol_table.cpp
```diff
@@ -0,0 +1,94 @@
...
+		if (insert_result.second) // introduced a previously undeclared name
+			return nullptr;
+		return std::exchange(insert_result.first->second, decl);
+	}();
+
+	auto insert_result =
+	    latest_shadowed_scope().insert({decl->identifier_text(), old_decl});
+	if (!insert_result.second)
+		Log::fatal() << "Redeclaration of '" << decl->identifier_text() << "'";
+}
```
lushoBarlett, line 30: I need help to understand this

lushoBarlett (approved):
I'll approve just in case it's all correct, but please explain the thing I don't understand first.

src/symbol_table.cpp
SebastianMestre: you mean the construct itself? It is an immediately invoked lambda, it works not unlike a seq-expr in Jasper itself.

As for the logic of it: we try to insert a new declaration.

If we succeed, we know the name was previously undeclared, so there is no previous declaration, so we return nullptr.

If we fail, we get an iterator that points to the previous declaration, so we assign through that iterator, and return the old value. (which is what `std::exchange` does)

Here is a pseudocode for `std::exchange`
```c++
template<typename T>
T exchange(T& dst, T&& new_value) {
    T old_value = std::move(dst);
    dst = std::move(new_value);
    return old_value;
}
```

lushoBarlett: Ahh, I see. Yea that makes sense.

PR #256, opened 2021-03-31
Better parser api

SebastianMestre:
So, I went and improved both the parser's and lexer's api, fully baking in the assumption that we first fully tokenize a file, and then parse it.

It is now a lot more value-oriented, and const-friendly, which aligns more with modern C++ values, and reads more nicely. (also, purely coincidentally, it is more SOLID compliant)

old style:
```c++
TokenArray ta;
CST::Allocator ca;
auto cst_handle = parse_program(source, ta, ca);
```

new style
```c++
TokenArray const ta = tokenize(source);
CST::Allocator ca;
auto cst_handle = parse_program(ta, ca);
```

Furthermore, the implementation of `parse_program` et al itself became a lot nicer (though that's mostly because it has a single responsibility now)

Other than that, I also moved the `Lexer` type definition into the module implementation file, which I feel is a lot tidier.

### Notes for reviewer

I recommend to read this PR commit by commit, as each one is a small refactoring (in the proper sense of the word).

### DIFF
--- a/src/interpreter/main.cpp
+++ b/src/interpreter/main.cpp
@@ -45,7 +45,7 @@ int main(int argc, char** argv) {
 		    // TODO: We need to clean this up
 
 		    {
-			    TokenArray ta = tokenize("__invoke()");
+			    TokenArray const ta = tokenize("__invoke()");
 
 			    CST::Allocator cst_allocator;
 			    auto top_level_call_ast = parse_expression(ta, cst_allocator);
--- a/src/interpreter/main.cpp
+++ b/src/interpreter/main.cpp
@@ -44,12 +44,12 @@ int main(int argc, char** argv) {
 		    // TODO: We need to clean this up
 
 		    {
-			    TokenArray ta;
-			    CST::Allocator cst_allocator;
-			    AST::Allocator ast_allocator;
+			    TokenArray ta = tokenize("__invoke()");
 
-				tokenize("__invoke()", ta);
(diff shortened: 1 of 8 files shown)


src/parser.cpp
```diff
@@ -15,6 +15,11 @@
 
 #include <cassert>
 
+static Token const& eof() {
+	static Token t = {TokenTag::END, "(EOF)", -1, -1, -1, -1};
+	return t;
+}
+
```
lushoBarlett, line 19: This is intentionally duplicated I presume. Not a big deal but to keep in mind.

src/interpreter/main.cpp
```diff
@@ -48,7 +48,8 @@ int main(int argc, char** argv) {
 			    CST::Allocator cst_allocator;
 			    AST::Allocator ast_allocator;
 
-			    auto top_level_call_ast = parse_expression("__invoke()", ta, cst_allocator);
+				tokenize("__invoke()", ta);
+			    auto top_level_call_ast = parse_expression(ta, cst_allocator);
```
lushoBarlett: bad indentation?

lushoBarlett (approved):
Lexer class has been hidden, and now the world is ruled by the tokenize function. I like this style.

src/parser.cpp
SebastianMestre: Good catch. I forgot to delete the old one after doing the refactoring

PR #257, opened 2021-04-12
Top level functions

SebastianMestre:
I added some syntactic sugar for function declarations.

We may now write

```rust
fn add(a, b) {
    return a + b;
};
```

And it will desugar to

```rust
add := fn (a, b) => seq {
    return a + b;
};
```

Likewise, regular functions work, too:

```rust
fn id(x) => x;
```

desugars to

```rust
id := fn(x) => x;
```


SebastianMestre:
Actually, I'm not happy with the code duplication, so I'm setting this to draft until I factor it out.

SebastianMestre:
I did a bunch of not-so-related things here, and this will probably conflict quite a bit with #259, so I'm closing this, and re-doing it after that PR is merged

PR #259, opened 2021-04-14
Cst touchup

SebastianMestre:
## Data Layout

I changed up some of how the Cst is organized: I put the data that goes in declarations into its own helper struct called `DeclarationData`.
This struct is smaller than CST::Declaration by probably 16 bytes (due to it not having a tag and a virtual pointer), so it is lighter-weight when stored by value in other nodes. (e.g. FunctionLiteral, ForStatement)

There is now a type alias for function arguments. It's just more convenient than typing out `std::vector<DeclarationData>` every single time.

Also, tokens no longer store 4 integers with their source location information, but a brand new `SourceRange` struct. Hopefully this will come in handy in many different places

## Error reporting

I refactored a good deal of how we generate errors in the parser et al, while introducing some visual improvements. It's not night-and-day, but I think it's significant.

Here is an example of how some errors are generated in the parser:

```diff
There is a newer and nicer API to add sub-errors to Writer<T>
-	result.m_error.m_sub_errors.push_back(
-	    make_expected_error("a declaration", p0));
+	result.add_sub_error(make_expected_error("a declaration", p0));

We can pass TokenTag to make_expected_error now
-	return {make_expected_error(
-	    token_string[int(expected_type)], current_token)};
+	return {make_expected_error(expected_type, current_token)};

This error now gives a source location
-	result.m_error.m_sub_errors.push_back(
-	    {{"Found a terminator after a delimiter in an "
-	      "expression list "}});
+	result.add_sub_error(make_located_error(
+	    "Found trailing delimiter in expression list", p0));
```

The way `match_identifiers` generates its errors has been unified with how the parser does it.

```diff
-	return {
-	    "at line " + std::to_string(token->m_line0 + 1) +
-	    " : accessed undeclared identifier '" + ast->text().str() + "'"};
+	return make_located_error(
+	    "accessed undeclared identifier '" + ast->text().str() + "'",
+	    token->m_source_location.start);
```

Whether this snippet of code is better is debatable, but I think it's definitely clearer, as it directly states its intent. This change also makes the error look more similar to the parser's

```diff
	- While scanning top level declaration '__invoke'
-	-- at line 7 : accessed undeclared identifier 'a'
+	-- At 7:3 : accessed undeclared identifier 'a'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ COMMON_TARGETS := \
 	match_identifiers \
 	metacheck \
 	parser \
+	source_location \
 	symbol_table \
 	token \
 	typecheck \
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -14,11 +14,10 @@ InternedString const& Declaration::identifier_text() const {
 			Log::fatal() << "No identifier string or fallback on declaration";
 
 		auto cst = static_cast<CST::Declaration*>(m_cst);
-		auto token = cst->m_identifier_token;
 
-		Log::warning() << "No identifier on declaration " << token->m_text << ": using token data as fallback";
+		Log::warning() << "No identifier on declaration, using token data as fallback: '" << cst->identifier() << "'";
 
-		return token->m_text;
+		return cst->identifier();
 	}
 
 	return m_identifier;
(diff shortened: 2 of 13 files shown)


lushoBarlett (approved):
Looks good enough to me. I mean, I understand the reason behind every change that made it to the final code. It's a good start to some other changes that I imagine can be made.

PR #260, opened 2021-04-21
Function declarations

SebastianMestre:
This is the reboot of #257. It came out to a smaller diff, though. (but with a greater amount of added lines)

I added some syntactic sugar for function declarations.

We may now write

```rust
fn id(x) => x;

// desugars to
id := fn(x) => x;
```

Likewise, block functions work, too:

```rust
fn add(a, b) {
    return a + b;
};

// desugars to
add := fn (a, b) => seq {
    return a + b;
};
```

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -30,7 +30,7 @@ Token const* Identifier::token() const {
 }
 
 
-SequenceExpression* convert_and_wrap_block_in_seq_expr(CST::Block* cst, Allocator& alloc) {
+SequenceExpression* convert_and_wrap_in_seq(CST::Block* cst, Allocator& alloc) {
 	auto block = static_cast<Block*>(convert_ast(cst, alloc));
 	auto seq_expr = alloc.make<SequenceExpression>();
 	seq_expr->m_body = block;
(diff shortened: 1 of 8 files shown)


src/cst.hpp
```diff
@@ -34,19 +37,67 @@ struct DeclarationData {
 	}
 };
 
+using FuncArguments = std::vector<DeclarationData>;
+
 struct Declaration : public CST {
+	// This function is very cold -- it's ok to use virtuals
+	virtual InternedString const& identifier_virtual() const = 0;
```
lushoBarlett, line 44: Doubt it will ever be a bottleneck, and since this is for parsing purposes it kinda does not matter.

src/ast.cpp
```diff
@@ -27,6 +29,15 @@ Token const* Identifier::token() const {
 	return static_cast<CST::Identifier*>(m_cst)->m_token;
 }
 
+
+SequenceExpression* convert_and_wrap_block_in_seq_expr(CST::Block* cst, Allocator& alloc) {
```
lushoBarlett, line 33: Expressive name. Perhaps `convert_and_wrap_in_seq` is good enough?

lushoBarlett (approved):
Very good quality-of-life addition right here!

src/ast.cpp
SebastianMestre: Yeah, good call

PR #264, opened 2021-04-28
push missing pizza expression arguments

### LINKED ISSUE #261: Pizza expressions are broken for functions that take more than one argument
While both of these pieces of code work perfectly:

```rust
f := fn(a, b) => 0;
__invoke := fn() => f(0, 0);
```

```rust
f := fn(a) => 0;
__invoke := fn() => 0 |> f(0);
```

This piece of code fails to typecheck:

```rust
f := fn(a, b) => 0;
__invoke := fn() => 0 |> f(0);
```

The error is:

```
jasperi: src/./algorithms/unification.cpp:65: void Unification::Core::unify(int, int): Assertion `term_data[vi].argument_idx.size() == term_data[vj].argument_idx.size()' failed.
```

NOTE: A lot of assertions in the typechecker should be fatal errors instead...

SebastianMestre:
fixes #261


lushoBarlett (approved):
Well that was a dumb mistake, wasn't it? Proof that we don't use pipe/pizza often.

PR #266, opened 2021-05-01
Value restriction

### LINKED ISSUE #265: Unsound type inference on mutable types
This piece of code typechecks.

```rust
fn __invoke() {
  // infer f : forall a. () -> array<:a:>
  f := fn() => array{};

  // instanciate f, then generalize, infer arr : forall b. array<:b:>... wth???
  arr := f();

  // instanciate arr with b=int<::>, infer a1 : array<:int<::>:>
  a1 : array<:int<::>:> = arr;

   // instanciate arr with b=string<::>, infer a2 : array<:string<::>:>
  a2 : array<:string<::>:> = arr;

  array_append(a1, 10); // ok, a1 is array of int
  array_append(a2, ""); // ok, a2 is array of string

  // arr := array { 10; ""; };

  return 0;
};
```

There is a well known and simple solution to this, I'm working on it.

SebastianMestre:
closes #265 

## Introduction

In ML-style type systems with mutable data, there is a distinction between 'values' and 'references'. A value is, essentially, an immutable value, while a reference is a handle to a mutable value.

To prevent the problems outlined in #265, we want to treat values and references differently in certain cases: we never want to generalize free vars that belong to a reference.

This idea gives us this truth table:

|has free vars?|value or reference?|should generalize var?|
|-----|--------|----|
|yes|value|yes|
|no|value|no|
|yes|reference|no|
|no|reference|no|

## The algorithm

A first stab at implementing this can be reasoned like so: we only want to generalize immutable things, and the only immutable thing that we have are functions. Thus, we will only generalize declarations of function expressions (i.e. things that look like `f := fn(...) => expr`)

This actually fixes #265, but is much too restrictive, as the following fails to typecheck:

```rust
fn __invoke() {
  id := fn(x) => x; // infer id: forall a. a -> a
  id2 := id;        // b:=newvar(), instanciate with a=b, then id2: b -> b (note the lack of forall)
  id2(0);           // unify b = int<::>
  id2("0");         // unify b = string<::> (gives a type error because b = int<::>)
};
```

To prevent this, we will also generalize identifier expressions.

At first glance, it might seem like things like the following will typecheck, (much like in the current master branch):

```rust
fn __invoke() {
  f := fn() => array{};
  a := f(); // has type array<:a:> (note lack of forall, due to new logic)

  // seems like it would generalize and have type forall b. array<:b:>, due to it being an identifier
  b := a;
  a1 : array<:int<::>:> = b;
  a2 : array<:string<::>:> = b;

  array_append(a1, 10);
  array_append(a2, "");

  return 0;
};
```

In reality, this will not happen, because the free variables of `a` will be part of the environment by the time `b` is typechecked, meaning they will not be generalized, by the usual means. (This is not implemented in master because we generalize every declaration, but now, when a declaration is not generalized, its free vars will get bound to the environment)

All in all, what we are doing is creating an explicit check for column 2 of the truth table. Well we actually overestimate, and end up taking up some of the false cases, too. In particular, that means we don't generalize some things that fall into row 1.

```rust
// has type forall a. a -> a
id := fn(x) => x;
// id2 is not a function expression, nor an identifier expression, so it won't be generalized
// has type b -> b (no forall)
id2 := id(id);
// we can get around this by using eta-expansion (i.e write the function explicitly)
// has type forall c. c -> c
id2 := fn(x) => id(id)(x);
```

But i feel like it's a low price to pay, and the workaround is easy enough. The overall technique is called **"the value restriction"**, and it's quite common in ML-style languages with mutation. There are other techniques that get some (or a lot) of row 1 back, but they're a lot more complex to implement.

https://en.wikipedia.org/wiki/Value_restriction

http://mlton.org/ValueRestriction

--------------------

PS: I sneaked in a tiny refactoring as well

PSS: since this makes the type system strictly more restrictive (i.e programs that previously typechecked no longer do) we don't have a way to do automated testing

### DIFF
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -11,18 +11,6 @@
 
 namespace TypeChecker {
 
-// this function implements 'the value restriction', a technique
-// that enables type inference on mutable datatypes
-static bool is_value_expression(AST::AST* ast) {
-	switch (ast->type()) {
-	case ASTTag::FunctionLiteral:
-	case ASTTag::Identifier:
-		return true;
-	default:
-		return false;
-	}
-}
-
 static void process_type_hint(AST::Declaration* ast, TypeChecker& tc);
 
 // Literals
@@ -289,6 +277,18 @@ void print_information(AST::Declaration* ast, TypeChecker& tc) {
 	Log::info("The type is:");
 	tc.m_core.m_mono_core.print_node(poly_data.base);
 #endif
+}
+
+// this function implements 'the value restriction', a technique
+// that enables type inference on mutable datatypes
+static bool is_value_expression(AST::AST* ast) {
+	switch (ast->type()) {
+	case ASTTag::FunctionLiteral:
+	case ASTTag::Identifier:
+		return true;
+	default:
+		return false;
+	}
 }
 
 void generalize(AST::Declaration* ast, TypeChecker& tc) {
--- a/src/typecheck.cpp
+++ b/src/typecheck.cpp
@@ -295,13 +295,17 @@ void generalize(AST::Declaration* ast, TypeChecker& tc) {
 	assert(!ast->m_is_polymorphic);
 
 	assert(ast->m_value);
-	if (!is_value_expression(ast->m_value))
-		return;
 
-	ast->m_is_polymorphic = true;
-	ast->m_decl_type = tc.generalize(ast->m_value_type);
+	if (is_value_expression(ast->m_value)) {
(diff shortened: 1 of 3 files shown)


lushoBarlett (approved):
Very interesting. I'm glad you found material to work this problem out in a simple fashion. I suggest we do Eta-Expansion as well. My suggestion would be to have a static function called `eta_expand` that just wraps a syntax. Or maybe there's something even simpler.

PR #267, opened 2021-05-09
Optimize the lexer

SebastianMestre:
I profiled and found that the frontend spent about 11% of time in the lexer.

### Optimization 1

A good chunk of that was spent inside `InternedString(char const*, size_t)`. Perf quickly showed me to see that it was mostly called from `Lexer::push_token`. In turn, that routine got a significant amount of calls from `Lexer::consume_symbol`. This is kinda silly: symbols are hard-coded into the language, and the lexer already knows which symbol was consumed when we push it, so there should be no need to do a hash-lookup on it, too.

To solve this, I stored `InternedString`s in the trie that we use to parse symbols. This way, each symbol gets looked up in the strings table only once: when we first initialize the lexer.

Turns out, most tokens are symbols in a typical Jasper program, so saving one hash-lookup per symbol is a pretty big deal.

### Optimization 2

Similarly to symbols, keywords are hard coded into the language, so we shouldn't need to do a hash lookup when we consume one.

We now put the possible keyword string into a `string_view` to compare it with the language's keywords, and take the `InternedString` from the keywords array, if we happen to match against one.

Although we now need to do string comparisons instead of pointer comparisons in that loop, because of the distribution of prefixes in the keywords (in particular, no two keywords have the same first and second letter as each other), the string comparisons are O(1).

> Side note: In my benchmark, the `InternedString` constructor spent 97% of time in `StringSet::scan`, so I tried to speed up that routine, but it didn't go too well. Instead, looking at where it was being used proved much more effective.

### Results

Tokenization used to take 11% of time, it now takes 9.9%. That would imply a speedup of 1.22% in the frontend, overall. Empirically, I measured a speedup of 1.8%, which is pretty close.

### DIFF
--- a/src/algorithms/trie.hpp
+++ b/src/algorithms/trie.hpp
@@ -4,12 +4,13 @@
 #include <cstdint>
 
 #include "../utils/string_view.hpp"
+#include "../utils/interned_string.hpp"
 
 static constexpr uint8_t nil = ~uint8_t(0);
 
 struct Trie {
 	struct Entry {
-		string_view text;
+		InternedString text;
 		int data;
 	};
 
--- a/src/lexer.cpp
+++ b/src/lexer.cpp
@@ -35,6 +35,7 @@ struct Lexer {
 	bool consume_identifier_or_keyword();
 	bool consume_number();
 	void consume_token();
+	void push_token(TokenTag, InternedString);
 	void push_token(TokenTag, int);
 
 	std::pair<bool, TokenTag> is_keyword(InternedString const&);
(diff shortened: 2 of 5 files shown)


SebastianMestre:
I did measurements in the tuning build, and those were, indeed, the results. However, they don't seem to translate well to the release build, where I measured a very slight (~0.1%) performance regression (?)

I'll do some more thorough benchmarks later.

SebastianMestre:
I'm closing this in favor of #270

PR #270, opened 2021-05-16
implement a table-based lexer

SebastianMestre:
### TODO

 - [x] Fill out the lexer table with all token types
 - [x] Use the correct tag for non-keyword non-symbol non-identifier tokens
 - [x] Push EOF token
 - [x] Integrate the new lexer
 - [x] Delete the old lexer
 - [x] Delete code related to the old lexer (`utils/trie`, etc)

-------------

Still WIP, however, results are quite promising. Here are some measurements from a "micro" benchmark, all taken using release compilation settings:

version | mean [ms] | stdev [ms]
----|-----|------
hand written | 9.133 | 0.470
naive | 6.643 | 0.543
fast | 5.560 | 0.359
speed of light | 2.377 | 0.281

The hand written version is the lexer in the master branch.

The naive version is a table based lexer.

The fast version is a table based lexer with the optimization of not repeatedly hashing and looking up `InternedString`s for symbols and keywords.

The "speed of light" (SOL) version loads a file from disk, and computes xor of all its bytes. It represents a baseline for the fastest possible implementation.

### Conclusion

The optimized table based lexer is about halfway between the current one and the SOL version. One could say it is about twice as fast.


SebastianMestre:
Re-measurements, after completing implementation:

version | mean [ms] | stdev [ms]
----------|----------------|----------------
hand written | 0.009316 | 0.000714
fast | 0.006182 | 0.000482
speed of light | 0.002667 | 0.000346

The new implementation got a bit slower, just due to it doing more things (like keeping track of source locations), but it's still a solid 33% faster.

src/automaton.cpp
```diff
@@ -0,0 +1,484 @@
...
+	// Identifiers
+	State saw_id_char = new_state();
+	add_transition(result, start, '_', saw_id_char);
+	for (char c = 'a'; c <= 'z'; ++c) add_transition(result, start, c, saw_id_char);
+	for (char c = 'A'; c <= 'Z'; ++c) add_transition(result, start, c, saw_id_char);
+	add_default_transition(result, saw_id_char, Identifier);
+	add_transition(result, saw_id_char, '_', saw_id_char);
+	for (char c = 'a'; c <= 'z'; ++c) add_transition(result, saw_id_char, c, saw_id_char);
+	for (char c = 'A'; c <= 'Z'; ++c) add_transition(result, saw_id_char, c, saw_id_char);
+	for (char c = '0'; c <= '9'; ++c) add_transition(result, saw_id_char, c, saw_id_char);
```
lushoBarlett, line 140: `RANGE('a', 'z')` macro is great for this. `add_self_transition(state, char)` is very helpful for readability.

src/automaton.cpp
```diff
@@ -0,0 +1,484 @@
...
+	// Implementation detail: We store indices into the source buffer
+	// in the source_locations at first, then resolve them in another pass.
+
+	char const* const code_start = p;
+
+	constexpr Automaton a = MainLexer::make();
+	constexpr Automaton ka = KeywordLexer::make();
+
+	TokenArray ta;
+	while (*p == ' ' || *p == '\t' || *p == '\n') ++p;
```
lushoBarlett, line 401: Used twice, low level and hard to look at. Free macro / lambda

src/automaton.cpp
```diff
@@ -0,0 +1,484 @@
...
+			int code_idx2 = token.m_source_location.end.line;
+
+			for (; code_idx < code_idx1; ++code_idx) {
+				if (code_start[code_idx] == '\n') {
+					line += 1;
+					col = 0;
+				} else {
+					col += 1;
+				}
+			}
```
lushoBarlett, line 464: This loop is also used twice.

lushoBarlett (approved):
Since you've written quite a lot, and it works, I'm fine with approving and merging this. Although, I must clarify, there's easy refactorings to be made. I'll do them myself sometime this week.

src/automaton.cpp
```diff
@@ -0,0 +1,484 @@
...
+#undef X
+constexpr int last_fixed_string = RPoly;
+}
+#define X(name, token_tag, string) string,
+static InternedString fixed_strings[] = { END_STATES };
+#undef X
+#define X(name, token_tag, string) TokenTag::token_tag,
+constexpr TokenTag token_tags[] = { END_STATES };
+#undef X
+#undef END_STATES
```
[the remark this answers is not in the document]
SebastianMestre: good point

src/automaton.cpp
SebastianMestre: I'm not sure about RANGE macro. I like the helper, I'm going with `add_loop`, though.

src/automaton.cpp
SebastianMestre: got it

SebastianMestre:
> Since you've written quite a lot ...

500 LOC is about the upper bound for a PR, IMO. So yeah, it is pretty much borderline on the size. Sorry about that.

> ... there's easy refactorings to be made. I'll do them myself sometime this week.

Cool! Looking forward to it!

src/automaton.cpp
SebastianMestre: fo sho

PR #272, opened 2021-05-23
match_identifiers in eval_expression

SebastianMestre:
I did some plumbing so that eval_expression can call match_identifiers. This means that we will sometimes get a nicer error if a test fails.


lushoBarlett (approved):
I can agree to this, there might be something to be done with the `Interpreter` and `context` but it's no big deal for now.

PR #273, opened 2021-05-23
Better xmacros

SebastianMestre:
While making the new lexer, I noticed we can use X-macros with multiple fields. So, I figured I would apply this newfound knowledge to the codebase.

I also renamed `ExitStatusTag` to `ExitStatus`, as it is not a tag (in the sense that we don't use it to discriminate between members of a sum type). Same thing for `TestStatusTag`


lushoBarlett (approved):
Goodie.

PR #274, opened 2021-05-27
Simplify lexer

lushoBarlett:
Small tweak to see if I can simplify the lexer a bit, and as much as I can

### DIFF
--- a/src/lexer.cpp
+++ b/src/lexer.cpp
@@ -107,45 +107,30 @@ constexpr void init_transitions(AutomatonBuilder& builder) {
 		builder.default_transition(i, Error);
 
 	State start = builder.new_state();
-	State saw_id_char = builder.new_state();
 	State saw_open_string = builder.new_state();
-	State saw_slash = builder.new_state();
 	State saw_comment_marker = builder.new_state();
-	State saw_digit = builder.new_state();
 	State saw_number_dot = builder.new_state();
-	State saw_decimal_digit = builder.new_state();
-	State saw_eq = builder.new_state();
-	State saw_bang = builder.new_state();
-	State saw_colon = builder.new_state();
-	State saw_lt = builder.new_state();
-	State saw_gt = builder.new_state();
-	State saw_plus = builder.new_state();
-	State saw_dash = builder.new_state();
-	State saw_star = builder.new_state();
-	State saw_pipe = builder.new_state();
-	State saw_amp = builder.new_state();
-	State saw_caret = builder.new_state();
 
 	builder
 
 	.from_state(start)
-		.transition('_', saw_id_char)
-		.transition(Range {'a', 'z'}, saw_id_char)
-		.transition(Range {'A', 'Z'}, saw_id_char)
+		.transition('_', Identifier)
+		.transition(Range {'a', 'z'}, Identifier)
+		.transition(Range {'A', 'Z'}, Identifier)
 		.transition('"', saw_open_string)
-		.transition(Range {'0', '9'}, saw_digit)
-		.transition('/', saw_slash)
-		.transition('=', saw_eq)
-		.transition('!', saw_bang)
-		.transition(':', saw_colon)
-		.transition('<', saw_lt)
-		.transition('>', saw_gt)
-		.transition('+', saw_plus)
-		.transition('-', saw_dash)
-		.transition('*', saw_star)
-		.transition('|', saw_pipe)
-		.transition('&', saw_amp)
-		.transition('^', saw_caret)
+		.transition(Range {'0', '9'}, Integer)
+		.transition('/', Slash)
+		.transition('=', Assign)
+		.transition('!', Bang)
+		.transition(':', Colon)
+		.transition('<', Lt)
+		.transition('>', Gt)
+		.transition('+', Plus)
+		.transition('-', Minus)
+		.transition('*', Star)
+		.transition('|', Pipe)
+		.transition('&', Amp)
+		.transition('^', Caret)
 	// simple single char tokens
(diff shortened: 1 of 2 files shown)


lushoBarlett:
Ending it here because it's not even worth it to further work in it. The process is now clear and without weirdness.

SebastianMestre:
I want benchmarks

src/lexer.cpp
```diff
@@ -412,16 +343,18 @@ TokenArray tokenize(char const* p) {
 		char const* const p0 = p;
 
 		int state = state_count - 1;
-		while (MainLexer::EndStates::Count < state) {
-			state = a.go(state, *p++);
+		int new_state = a.go(state, *p++);
+		while (new_state != MainLexer::EndStates::Error) {
+			state = new_state;
+			new_state = a.go(state, *p++);
 		}
```
SebastianMestre, line 351: I think I prefer this expressed as an infinite loop with an if-break, to eliminate the need of writing the `new_state` assignment twice.

(if your thing is faster, though, let's use that)

```suggestion
		while(1) {
			int new_state = a.go(state, *p++);
			if (new_state == MainLexer::EndStates::Error) break;
			state = new_state;
		}
```

src/lexer.cpp
```diff
@@ -412,16 +343,18 @@ TokenArray tokenize(char const* p) {
...
 
-		if (state == MainLexer::EndStates::Error) {
+		if (MainLexer::EndStates::Count <= state) {
 			print_error(p);
 			break;
 		}
```
SebastianMestre, line 355: I'm struggling with this logic, tell me if i understood correctly:

since we always stop before reaching the error state, an error manifests as stopping in a non-end state?

SebastianMestre (commented):
It looks like it does one more iteration for multi-char tokens, so I'm concerned performance could suffer. I would like to see some performance measurements before accepting this.

Other than that, it's very nice. It's always good when something can be written with fewer lines and fewer concepts.

src/lexer.cpp
SebastianMestre: i looked at the asm output, and they both generate the exact same code for the inner loop, but the surrounding code is a bit longer for yours.

There is also this version, which generates a weird instruction I don't know in the inner loop, but otherwise is very similar to my other sugestion

```suggestion
		while (1) {
			int new_state = a.go(state, *p++);
			if (new_state == MainLexer::EndStates::Error) break;
			state = new_state;
		}
```

SebastianMestre: All of these (including yours) have 2 more instructions in the inner loop (for a total of 10) than the original (which has 8)

src/lexer.cpp
lushoBarlett: We stop when the next step is an error state. So if we are not in an end state when that happens, then the thing is clearly wrong. So yes, you got that right.

lushoBarlett:
Perf results just dropped! And the results are interesting.

I did 3 attempts for both methods. First and second attempt used `500 KLOC` and third attempt used 4 MLOC.

Each attempt used this pattern
```rust
test := fn() {
    t := false;
    
    t = seq {
        a := "hello";
        b : int<::> = 1;
        c : float<::> = 23.01;
        d : array {1; 2; 3; 4;};
        z : boolean<::> = false;
        if (1 < 1) {
            e := fn() => true;
            z = e();
        } else {
            f := fn() => false;
            z = f();
        }
        return z;
    };
    .
    .
    .
    
    return t;
    }
};

__invoke = fn() => 0;
```
Command used for every attempt was
```perf stat -d -r 100 --table -o output ./bin/jasperi lexer.jp```
I have the files if you want more details on each run.

## The runs

| | master | refactor |
| | --- | --- |
| `500 KLOC` 1 | `1.06517 +- 0.0127` | `1.063717 +- 0.000870` |
| `500 KLOC` 2 | `1.067041 +- 0.000739` | `1.063931 +- 0.000837` |
| `4 MLOC` | `8.75898 +- 0.00573` | `8.71574 +- 0.00743` |

SebastianMestre (approved):
I measured myself, failed to reproduce your results: I found no significant difference in speed. (I did observe that your version takes fewer branches, though, which seems mysterious to me)

Since this is clearly better code, and there is no significant drop in performance, I am accepting it.

SebastianMestre:
Btw the benchmark I used looks like this:

```c++
#include "src/token.hpp"
#include "src/lexer.hpp"

constexpr char const* source = R"(

edge := struct {
	u : int<::>;
	v : int<::>;
	w : int<::>;
}<::>;

init := fn(n) { result := array{}; for (i := 0; i < n; i = i+1) result |> array_append(i); return result; };
join := fn(uf, i, j) { uf[find(uf, i)] = find(uf, j); return null; };
find := fn(uf, i) { if (uf[i] == i) return i; return uf[i] = find(uf, uf[i]); };
sort := fn(arr, cmp) { for (i := 1; i < size(arr); i = i+1) { for (j := i; j < size(arr); j = j+1) { if (cmp(arr[j], arr[j-1])) { t := arr[j]; arr[j] = arr[j-1]; arr[j-1] = t; } } } return arr; }; 
kruskal := fn(edges : array<:edge:>, n) { uf := init(n); sort(edges, fn(a, b) => a.w < b.w); result := array{}; for (i := 0; i < size(edges); i = i+1) { if (find(uf, edges[i].u) == find(uf, edges[i].v)) { } else { result |> array_append(edges[i]); } } return result; };

// several thousand lines follow ...

fn __invoke() {
	return 0;
};

)";

int main () {
	auto tokens = tokenize(source);
	return tokens.size();
}

PR #276, opened 2021-06-06
Put scalar values directly on the stack

SebastianMestre:
The overview is:

First, use `Handle`, a new struct type, instead of `Value*` to refer to values. Initially, this was just a wrapper around a `Value*`.

Then, I put the tag in the `Handle`, and followed it up with with adding a union that holds either a pointer to the heap, or a boolean which is stored directly on the `Handle`.

Intend on doing this transformation for floats and integers, too.

### DIFF
--- a/src/interpreter/eval.cpp
+++ b/src/interpreter/eval.cpp
@@ -229,10 +229,10 @@ void eval(AST::TernaryExpression* ast, Interpreter& e) {
 	// TODO: proper error handling
 
 	eval(ast->m_condition, e);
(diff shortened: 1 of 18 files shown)


lushoBarlett (approved):
I found no blunders of logic, and I see you've actually fixed some bugs. I like the `Handle` interface, the fact that it is a value and not a pointer or reference in the codebase, the way we create it and use its values. Very welcome addition to the family, I expect great things from it!

PR #277, opened 2021-06-06
More data unboxing

SebastianMestre:
I put integers and numbers directly inside handles.

Comparing with before #276, I measured a 40% speedup on a benchmark that's heavy on integer math and comparisons (essentially best case scenario for this optimization), and 19% in another one that does some function calls and other stuff on the side.

I also fixed a few bugs from my previous pr along the way, and removed some outdated interfaces.


lushoBarlett (approved):
Very cool speedup and code simplification.

PR #278, opened 2021-06-08
WIP new metatype unifier

SebastianMestre:
Long term, I would like to merge `metacheck`, `ct_eval`, and `typecheck`. Into one or, at most, two passes.

For this purpose, I'm building a very custom unifier, which will be able to do metacheck in a single pass.

The overall design is that we store information about metatype relationships in a unification data structure, and only once they are all established, do we ever try to figure out exact metatypes.

This amount to: Instead of branching on the record metatype to figure out the metatype of the accessed element, we move that logic into the data structure, and have a node that represents that relationship. Then, in a later pass (currently, I plan on doing this in ct_eval), once all the relationships are established, we can resolve the actual metatypes (I think a good word for this step could be "materialize").

I am confident that this design will let us add types later on. It should be less glitchy and provide better error messages, too.


SebastianMestre:
Unifier might not be a great name: it doesn't have a `unify` method.

The usage is roughly the same as it is now. It pretty much would look like this:

```c++
// in metacheck.cpp
void metacheck(MetaUnifier& uf, AST::CallExpression* ast) {
  ast->m_metatype = uf.create_const_node(Tag::Value);

  metacheck(uf, ast->m_target);
  uf.turn_into(ast->m_target->m_metatype, Tag::Value);

  for (auto& arg : ast->m_args) {
    metacheck(uf, arg);
    uf.turn_into(arg->metatype, Tag::Value);
  }
}

void metacheck(MetaUnifier& uf, AccessExpression* ast) {
  metacheck(uf, ast->m_target);
  ast->m_metatype = uf.create_dot_node(ast->m_target->m_metatype);
}

// in ct_eval.cpp

AST::CallExpression* ct_eval(MetaUnifier& uf, AST::CallExpression* ast) {

  ast->m_target = ct_eval(uf, ast->m_target);

  for (auto& arg : ast->m_args)
    arg = ct_eval(uf, arg);

  return ast;
}

AST::AST* ct_eval(MetaUnifier& uf, AccessExpression* ast) {
  auto metatype = uf.eval(ast->m_metatype);

  if (uf.is(metatype, Tag::DotResult))
    Log::fatal() << "failed to deduce metatype for access expression";

  if (uf.is(metatype, Tag::Ctor))
    return ctor_from_ast(uf, ast);

  if (uf.is(metatype, Tag::Value)) {
    ct_eval(uf, ast->m_target);
    return ast;
  }

  Log::fatal() << "(weird) fully deduced invalid metatype for access expression";
}
```

src/ct_eval.cpp
```diff
@@ -323,19 +336,28 @@ AST::DeclarationList* ct_eval(
...
-					static_cast<AST::TypeFunctionHandle*>(decl->m_value);
+			int meta_type = uf.eval(decl->m_meta_type);
 
+#if 0
+			std::cerr << "[ Pass 2 ] top level \"" << decl->m_identifier << "\"\n";
+			std::cerr << "           metatype tag is: Tag(" << int(uf.tag(meta_type)) << ")\n";
+#endif
+
+			if (uf.is(meta_type, Tag::Var))
+				Log::fatal() << "Incomplete metatype inference on top level variable \"" << decl->m_identifier << "\"";
```
SebastianMestre, line 347: this makes me happy

lushoBarlett (approved):
This was reviewed live. Nice work.

PR #283, opened 2021-10-22
refine ambiguous nodes after metacheck

SebastianMestre:


SebastianMestre:
> ```c++
> - assert(target->type() == ASTTag::UnionAccessExpression); 
> - auto access = static_cast<AST::UnionAccessExpression*>(target);
> + auto access = sanity_cast<AST::UnionAccessExpression*>(target);
> ```
> 
> Actually quite useful everywhere, I don't know why we didn't think of that.

I thought of that, but it requires some template metaprogramming, that I was hesitant to add. Essentially, you need a way to map the type `AST::UnionAccessExpression` to the enum value `ASTTag::UnionAccessExpression`, which implies something like this:

```c++
template<typename T>
struct ast_trait {};

template<>
struct ast_trait<AST::UnionAccessExpression> {
    static constexpr ASTTag tag_value = ASTTag::UnionAccessExpression;
};
// ... repeat for every AST node type ...

template<typename T>
T* ast_cast(AST::AST* ast) {
  assert(ast->type() == ast_trait<T>::tag_value);
  return static_cast<T>(ast);
}
```

This is similar to what we have in the interpreter for casting heap values. The trait classes can probably be generated using the X-macros we already have. Will think about it.

> ```c++
> std::unordered_map<InternedString, MonoId> structure; 
> structure[access->m_member] = tc.new_var();  // isn't there an unordered_map constructor that does this?
> ```

Probably, yeah, good call.

> ```c++
> MonoId dummy_monotype =
>     tc.m_core.new_term(dummy_tf, {}, "Union Constructor Access"); // inconsistent line breaks, oh well
> ```

It's what clang-format chose IIRC.

PR #284, opened 2021-10-26
Constrained typevars

SebastianMestre:
I have identified and addressed two pain points in the type system implementation.

## `Unification::Core` is an incidental data structure

The union-find data structure is intertwined with the data structure of types. This forces one to deal with both at the same time, every time.

The fix for this was coding up a simple-as-can-be union-find implementation, that will keep track of variables. The types are now kept in a separate array, and there is no more funny 'turning apps into vars' business.

## Putting constraints only on type functions is the wrong abstraction

I noticed the very obvious pattern of creating a new dummy type function, then applying it a single time to get, essentially, a constrained monotype var.

I decided to instead have monotype vars hold constraints, too. A few things became easier, and I don't think anything got more complicated because of this change.

A functional consequence of this change is that we can now generalize over constrained types. I did not fix `inst_fresh` or `gather_free_vars` to handle this correctly, so constraints are removed during instantiation, which is wrong (e.g. a polymorphic `fn(x)=>x.a` can be called with a type that doesn't have an `a` field, and it typechecks).

--------------------------

PS: I removed all references to tuples from the codebase, because they'd probably never get finished anyways.


SebastianMestre:
I think these changes are a good idea, but maybe not quite in this form. I will give this a second go later, perhaps in two steps.

PR #285, opened 2022-01-23
Wrap symbol resolution code in a class

SebastianMestre:
This patch will:

 - rename `match_identifiers` to `resolve_symbols`, to be more in line with widely accepted terminology.
 - extract a few helper data structures from `SymbolResolutionHelper`
 - wrap symbol resolution code in a command class named `SymbolResolutionCommand`, making access and passing of helper data structures a lot more convenient.
 - Inline `SymbolResolutionHelper` into `SymbolResolutionCommand`

Another interpretation of the two last steps might be that the patch will

- rename `SymbolResolutionHelper` to `SymbolResolutionCommand`
- move the implementation of `match_identifers` into methods of `SymbolResolutionCommand`


lushoBarlett (approved):
I did not vibe at first with changing what's essentially a mapping from data to data into an object. Having it as a set of functions is sensible, so in that frame of mind the change seems unnecessary.

But then I realized that there's other objects involved, and having those in an object themselves removes a lot of argument passing into the methods of that object.

The other objects prove useful, I'm 50/50 on the one that's actually just a bunch of methods as it is inconsistent with the rest of the codebase.

It's just a style and each one has their own benefits.

Edit: the big one is technically one method that is overloaded

PR #286, opened 2022-01-29
Metacheck improvement

SebastianMestre:
This patch will change the paradigm of how meta type inference works behind the scenes.

The main idea is that, instead of trying to do inference by focusing on single variables, we take a more global view and try to solve all the constraints at the same time, in a single go. This makes things a lot simpler in my opinion.

The algorithm I used is very inefficient (it loops over all constraints until no further inferences can be made), but it is simple, maintainable, and easy to verify. It is also relatively easy to optimize: just add some hash tables to cut down on look ups.

PS: The commits are pretty good, so they can be looked at one-by-one

### DIFF
--- a/src/ct_eval.cpp
+++ b/src/ct_eval.cpp
@@ -112,7 +112,7 @@ static AST::Expr* ct_eval(
 	MetaTypeId meta_type = uf.eval(ast->m_meta_type);
 
 	// TODO: support vars
-	if (uf.is(meta_type, Tag::Ctor))
+	if (uf.is_ctor(meta_type))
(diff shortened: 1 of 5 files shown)


lushoBarlett (approved):
This is a very clean approach. Sure it might be a loss of time but one can never be sure until some measurements are made.

Some style decisions I don't fully agree with are
- do while loop is a better fit for the solve method I think
- not checked, but I think we don't favor "if one liners". I do prefer the branch action on a separate line myself.

The rest is good.

As a side note, a few "spoilers" managed to leak during git add --patch. It made me chuckle a bit :)

PR #287, opened 2022-01-29
eliminate tuples from the codebase

SebastianMestre:
Realistically, tuples will never be implemented in jasper, so might as well get rid of all the related code, since it's just dead weight.


lushoBarlett (approved):
Agreed

PR #288, opened 2022-01-30
Extract incidental union find typefunc unification logic

SebastianMestre:
This patch simplifies typefunc unification logic by making it no longer rely on the cumbersome `Unification::Core` class.

There is a deeper observation to be made about `Unification::Core` in general, which might be why it is so cumbersome to use and work on: It hides a union-find data structure inside of it. Part of this patch is making the data structure explicit, which simplifies reasoning.

It would be nice to do a similar thing for the other usage of `Unification::Core`, but that one may prove more difficult, as it does depend on all of the bells and whistles that the class includes.

Notice: This is a draft, but I'd still like a review. The core change is done, but the code was left in a less than ideal state due to a series of refactorings mean to facilitate this change. Any help in improving code quality is appreciated.

Notice: The commits are fairly self contained and cohesive, though some of the code was overwritten in the later ones.

### DIFF
--- a/src/algorithms/union_find.hpp
+++ b/src/algorithms/union_find.hpp
@@ -1,0 +1,28 @@
+#pragma once
+
+#include <vector>
+
+struct UnionFind {
+	int new_var() {
+		int result = repr.size();
+		repr.push_back(result);
+		return result;
+	}
+
+	int find(int i) {
+		if (repr[i] == i)
+			return i;
+		return repr[i] = find(repr[i]);
+	}
+
+	void join_left_to_right(int i, int j) {
+		repr[find(i)] = find(j);
+	}
+
+	bool is_single(int i) {
+		return i == repr[i];
+	}
+
+  private:
+	std::vector<int> repr;
+};
--- a/src/algorithms/union_find.hpp
+++ b/src/algorithms/union_find.hpp
@@ -3,7 +3,7 @@
 #include <vector>
 
 struct UnionFind {
-	int new_var() {
+	int new_node() {
 		int result = repr.size();
 		repr.push_back(result);
 		return result;
--- a/src/typesystem.cpp
+++ b/src/typesystem.cpp
@@ -180,19 +180,20 @@ void TypeSystemCore::unify_tf_data(TypeFunctionData& a_data, TypeFunctionData& b
 
 	b_data.argument_count = new_argument_count;
 
+	bool can_add_fields_to_b = b_data.is_dummy;
 	for (auto& kv_a : a_data.structure) {
-		auto kv_b = b_data.structure.find(kv_a.first);
+		auto const& field_name = kv_a.first;
+		auto const kv_b = b_data.structure.find(field_name);
 
-		if (kv_b == b_data.structure.end())
-			// if b doesn't have a field of a, act accordingly
-			if (b_data.is_dummy)
-				b_data.structure.insert(kv_a);
-			else
-				Log::fatal() << "Accessing non-existing field '" << kv_a.first << "' of a record";
-		else
-			// else the fields must have equivalent types
+		bool const b_has_field = kv_b != b_data.structure.end();
+		if (b_has_field)
 			m_mono_core.unify(kv_a.second, kv_b->second);
+		else if (can_add_fields_to_b)
+			b_data.structure.insert(kv_a);
+		else
+			Log::fatal() << "Accessing non-existing field '" << field_name << "' of a record";
 	}
+
 }
 
 TypeFunctionData& TypeSystemCore::get_tf_data(TypeFunctionId tf) {
--- a/src/typesystem.cpp
+++ b/src/typesystem.cpp
@@ -123,7 +123,7 @@ TypeFunctionId TypeSystemCore::create_tf(
     std::vector<InternedString> fields,
     std::unordered_map<InternedString, MonoId> structure,
     bool is_dummy) {
-	TypeFunctionId result = m_tf_uf.new_var();
+	TypeFunctionId result = m_tf_uf.new_node();
 	m_type_functions.push_back(
 	    {tag, arity, std::move(fields), std::move(structure), is_dummy});
 	return result;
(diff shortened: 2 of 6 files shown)


src/algorithms/union_find.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <vector>
+
+struct UnionFind {
+	int new_var() {
+		int result = repr.size();
+		repr.push_back(result);
+		return result;
+	}
```
lushoBarlett, line 10: The concept "new var" is related to the use cases of the union find, maybe a more generic name would be better, because we are trying to abstract the algorithm away from the codebase. My suggestions would be to keep "new" or maybe use "create" or "add". As for "var", maybe use "node".

src/algorithms/union_find.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+	int find(int i) {
+		if (repr[i] == i)
+			return i;
+		return repr[i] = find(repr[i]);
+	}
+
+	void join_left_to_right(int i, int j) {
+		repr[find(i)] = find(j);
+	}
```
lushoBarlett, line 20: I understand why we don't want to forget the direction of join, mainly because where we join stuff there's a "predominant" object (like a builtin function). In other words we need to choose the representative each time we join. Still, I'm not very happy with that long of a name in a well known algorithm that just calls this "join". I'm willing to keep it tho for the reason written above. Maybe "join_right" suffices?

src/algorithms/union_find.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+	bool is_single(int i) {
+		return i == repr[i];
+	}
```
lushoBarlett, line 24: Dead function, not used anywhere as of yet.

src/typesystem.hpp
```diff
@@ -67,4 +62,26 @@ struct TypeSystemCore {
...
+	TypeFunctionId create_tf(
+	    TypeFunctionTag tag,
+	    int arity,
+	    std::vector<InternedString> fields,
+	    std::unordered_map<InternedString, MonoId> structure,
+	    bool is_dummy);
+
+	std::vector<TypeFunctionData> m_type_functions;
+	std::vector<PolyData> poly_data;
+	UnionFind m_tf_uf;
```
lushoBarlett, line 86: Nomenclature is all over the place. We use "tf" and "type_function" interchangeably, we should pick one.

src/typesystem.cpp
```diff
@@ -191,3 +98,109 @@ void TypeSystemCore::gather_free_vars(MonoId mono, std::unordered_set<MonoId>& f
...
+
+		if (kv_b == b_data.structure.end())
+			// if b doesn't have a field of a, act accordingly
+			if (b_data.is_dummy)
+				b_data.structure.insert(kv_a);
+			else
+				Log::fatal() << "Accessing non-existing field '" << kv_a.first << "' of a record";
+		else
+			// else the fields must have equivalent types
+			m_mono_core.unify(kv_a.second, kv_b->second);
```
lushoBarlett, line 196: Not very important but I believe flattening this branch logic improves readability. You can choose to keep it like this.

src/typesystem.hpp
```diff
@@ -67,4 +62,26 @@ struct TypeSystemCore {
...
 	MonoId inst_fresh(PolyId poly);
+
+	TypeFunctionData& type_function_data_of(MonoId);
+	TypeFunctionId new_tf_var();
+	void unify_tf(TypeFunctionId, TypeFunctionId);
+private:
+	int compute_new_argument_count(
+	    TypeFunctionData const& a_data, TypeFunctionData const& b_data) const;
+
+	void point_tf_at_another(TypeFunctionId a, TypeFunctionId b);
```
lushoBarlett, line 73: Shouldn't this be called something like "unify_right"? I'm not very sure, I find the name to be too extravagant. I guess that _is_ the action that we want to perform at a high level so It's fine.

lushoBarlett (changes requested):
Refactorings look good, improvements to the structure and readability were made. You are right that it is still a bit messy, due to the code being old. Here are my thoughts.

src/typesystem.hpp
SebastianMestre: Yeah, I had some back and forth with this one, but in the end failed to come up with anything better :confused:

src/typesystem.cpp
SebastianMestre: aight I'll try some stuff out

src/typesystem.hpp
SebastianMestre: Yeah, which one do you like? Reading code, `tf` is immediately recognizable, but `type_function` is just nice to see innit? I used `tf` because it was less typing

src/algorithms/union_find.hpp
SebastianMestre: mm yes. I'd have to check, but I'm pretty sure I had used it for something in my other PR.

src/algorithms/union_find.hpp
SebastianMestre: I kinda like this one tbh

src/algorithms/union_find.hpp
SebastianMestre: yeah, good point. so `new_node`? Coming right up

I would like to start using the word `create` more often, as it evokes the picture that a value is being produced, but maybe some side effects happen as well.

src/typesystem.hpp
lushoBarlett: `type_function` is best in my opinion. The names of the functions need to be inversely proportional to the scope in which they live, as Uncle Bob stated.

src/typesystem.hpp
lushoBarlett: Ah that's okay, the name is explanatory.

src/typesystem.hpp
SebastianMestre: I don't think that's relevant

SebastianMestre:
I consider this change 'done', so I'd like a second look at it :^)

SebastianMestre:
Woah just noticed that the tests fail on CircleCI. Looks like something behaves differently on debug mode vs release mode. Will investigate

SebastianMestre:
This could use some refactoring. I feel like one or two classes could be extracted? Idk

PR #291, opened 2022-02-03
Refactor tc

SebastianMestre:
A decent chunk of the interface of the Typechecker class was used exclusively in `typecheck.cpp`, so I moved that code into its own class in that file.

I also moved the declaration order from `CompileTimeEnvironment` to `TypeChecker`


lushoBarlett (approved):
Great stuff! I like the method extractions and shuffles. Always welcome.

PR #292, opened 2022-02-03
Remove `interpreter/eval`'s dependency on the typechecker

SebastianMestre:


lushoBarlett (approved):
Wait what's this about? It doesn't break anything, so approving it is okay, but why is this possible?

SebastianMestre:
> Wait what's this about? It doesn't break anything, so approving it is okay, but why is this possible?

There's enough indirection that we can always get the things we need from the ast in this way

PR #293, opened 2022-02-10
Bugfix typevar in environment

SebastianMestre:
We used to only check if a variable was directly bound to the environment, but we actually need to recurse down all types that are bound to it, in case the variable shows up inside them

In terms of the Hindley-Milner system, (our implementation of) `\Bar{\Gamma}(\tau)` was correct, but `free(\Gamma)` was wrong.

The definition of `free(\Gamma)` can be seen in the figure in https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system#Syntax

The definition of `\Bar{\Gamma}(\tau)` can be seen in the figure in https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system#Syntax-directed_rule_system

### DIFF
--- a/src/typechecker/typecheck.cpp
+++ b/src/typechecker/typecheck.cpp
@@ -75,19 +75,13 @@ private:
 			return scope.m_type_vars.count(var) != 0;
 		};
 
-		// scan nested scopes from the inside out
-		for (int i = scopes().size(); i-- > 1;) {
+		// scan nested scopes from the inside out, including
+		// the global scope, at index 0
+		for (int i = scopes().size(); i-- ;) {
 			auto found = scan_scope(scopes()[i], var);
 			if (found)
 				return true;
-			if (!scopes()[i].m_nested)
-				break;
 		}
-
-		// fall back to global scope lookup
-		auto found = scan_scope(global_scope(), var);
-		if (found)
-			return true;
 
 		for (auto& scope : scopes()) {
 			for (auto stored_var : scope.m_type_vars) {
--- a/src/typechecker/typecheck.cpp
+++ b/src/typechecker/typecheck.cpp
@@ -85,8 +85,7 @@ private:
 
 		for (auto& scope : scopes()) {
 			for (auto stored_var : scope.m_type_vars) {
-				std::unordered_set<MonoId> free_vars;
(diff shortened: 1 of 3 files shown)


src/typechecker/typecheck.cpp
```diff
@@ -61,26 +61,43 @@ struct TypecheckHelper {
 	Frontend::CompileTimeEnvironment& env() { return tc.env(); }
 
 	TypeChecker& tc;
+
+	bool is_bound_to_env(MonoId var) {
+
+		for (auto& scope : env().scopes())
+			for (auto bound_type : scope.m_type_vars)
+				if (free_vars_of(bound_type).count(var) != 0)
+					return true;
```
lushoBarlett, line 70: This whole thing has mixed levels of care put into it.

First `env().scopes()` reads a little strange to me, you are using encapsulation yet it feels just like regular code. 

Second, `m_type_vars` is not encapsulated.

Lastly, the predicate in the if statement is "low level" logic and could be extracted into a function, e.g. `appears_free_in(bound_type, var)`

SebastianMestre: I'm not trying to hide the presence of `env`, only its location, because it will likely move into `TypecheckHelper` soon. Also, _it is_ regular code, wdym?

I don't think m_type_vars needs to be encapsulated: `Scope` is just a data holder. If anything should be encapsulated, it's scopes itself, but i haven't thought of a way to do that nicely yet.

SebastianMestre:
> Would've been nice to have a test first reveal the bug, then have this code fix it. Would be nice to have a test for this anyway, since it was a bug.

I did use a test when writing the patch. Here it is:

```
fn __invoke() {
  x := array{};
  g := fn () => x[0][0];
  array_append(array{3});
  z := array{ "z"; g(); };
  return 0;
}
```

The correct behavior is this code does not typecheck. If you figure out a way to write that test using our current testing setup, you unironically get a free cookie (or any dessert of your choice) :^).

src/typechecker/typecheck.cpp
lushoBarlett: Ah, so encapsulation helps with location. That's a fair point. I can understand the rest as well.

PR #294, opened 2022-02-15
Don't create useless `AST::*Handle` nodes

SebastianMestre:
When encountering a type func or a mono type, `ct_eval` essentially encodes the same data in a format that is suitable for the type checker. As it's doing that, though, it also wraps each type in a new AST node.

However, most of those nodes are discarded as soon as they are created (see `eval_then_get_mono` in the current master branch). For example, a declaration like `pt := struct { x: int<::>; y: int<::>;  }<::>;` causes three nodes to be created and promptly discarded. I think this is just silly.

Furthermore, all the wrapping and unwrapping makes the code slower and harder to read.

For these reasons, I tried to split the code that translates ASTs into typechecker data from the code that wraps the result.

I used really tiny refactorings to make these changes, so It may be annoying to read through the commit log. I am willing to condense the commit log, if that's something the reviewer is interested in.

### DIFF
--- a/src/typechecker/ct_eval.cpp
+++ b/src/typechecker/ct_eval.cpp
@@ -18,12 +18,12 @@ static void ct_visit(AST::Block* ast, TypeChecker& tc, AST::Allocator& alloc);
 
 static AST::Constructor* constructor_from_ast(AST::Expr* ast, TypeChecker& tc, AST::Allocator& alloc);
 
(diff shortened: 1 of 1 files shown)


SebastianMestre:
I'm going to condense the commit log into something more manageable then create another pull request. I think this one is just too much

PR #295, opened 2022-02-15
Don't create useless `AST::*Handle` nodes

SebastianMestre:
When encountering a type func or a mono type, `ct_eval` essentially encodes the same data in a format that is suitable for the type checker. As it's doing that, though, it also wraps each type in a new AST node.

However, most of those nodes are discarded as soon as they are created (see `eval_then_get_mono` in the current master branch). For example, a declaration like `pt := struct { x: int<::>; y: int<::>; }<::>;` causes three nodes to be created and promptly discarded. I think this is just silly.

Furthermore, all the wrapping and unwrapping makes the code slower and harder to read.

For these reasons, I tried to split the code that translates ASTs into typechecker data from the code that wraps the result.

> This was originally over 60 commits that I squashed down to 29. It's still a lot, but it's hard to crunch down further without diluting each commit.

### DIFF
--- a/src/typechecker/ct_eval.cpp
+++ b/src/typechecker/ct_eval.cpp
@@ -262,9 +262,19 @@ static MonoId compute_mono(
 		assert(arg);
 		assert(arg->type() == ASTTag::Identifier || arg->type() == ASTTag::TypeTerm);
 		if (arg->type() == ASTTag::Identifier) {
-			auto arg_handle = ct_eval(arg, tc, alloc);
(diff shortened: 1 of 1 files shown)


lushoBarlett (approved):
Good refactorings. Very orderly. Some instances where `for` and `if` braces could have been removed, no need to be so picky tho.

SebastianMestre:
> Good refactorings. Very orderly. Some instances where `for` and `if` braces could have been removed, no need to be so picky tho.

I'd be happy to hear any and all nitpicking, though!

PR #298, opened 2022-02-24
Remove handles

lushoBarlett:
Removes type function handles from the codebase, repurposes handles only for builtins.

PR #299, opened 2022-02-24
Remove mono handles

SebastianMestre:
Following the trend from #298, this PR removes creation of `MonoTypeHandle`s in ct_eval.

Turns out, we're not creating them anywhere else, so I removed them from the codebase.

~As a side note, I don't think we rewrite AST nodes in `ct_eval` anymore, so we could change its return type to void.~ We still rewrite some nodes into `AST::Constructor`s


lushoBarlett (approved):
LGTM

PR #306, opened 2022-03-15
Type constraints

SebastianMestre:
These are the main changes that happened here:

 - `Unification::Core` was inlined into `TypeSystemCore`. Turns out, there was a single instance of `Unification::Core` in the entire codebase, and it only complicated things.
 - The interface of `TypeSystemCore` was refined, preferring more use-case specific methods over more general methods that are more closely tied to the internal implementation.
 - The way type variables work was substantially changed. It now more closely resembles the theory, at the cost of some performance (I didn't measure, but there's a lot more indirection now, and there is more work being done)
 - Type constraints are no longer tracked indirectly through dummy typefuncs, but explicitly through a constraint object for each type variable. (This not entirely true, as we still need to constraint typefunc arities, but it's close enough)

As a side effect of this restructuring, I can tell that some soundness bugs were introduced, but not in any way that is detected by our current test suite. I think this is a necessary evil in order to improve the state of things.

### A new approach for type variables

Monotypes can have two forms: term and var. This corresponds to Haskell types like `Maybe Int` and `Maybe a`, where the argument to `Maybe` is a term and a var, respectively.

Each var has a `VarId` that identifies it, and the same var can conceptually appear multiple times, by having multiple var nodes with the same `VarId`. For instance, in the type `(a, [a])`, there would be two var nodes with the same `VarId`.

As we typecheck, two things can happen to a var:
 - It gets unified with a concrete type
 - It gets unified with another var

We used to have a single (really ugly) data structure to deal with these cases, but we now have two.

For the first case, we build a mapping from `VarId` to monotype (called `m_substitution` in the code). This mapping can only be established once. From then on, the var is treated as if it were the monotype itself, and no further actions may be taken upon the var itself. (The mental model is that the var happened to be a simple stand-in for some missing information, that has now been found out)

> `m_substitution` should be thought of as a function `substitution : Var -> Type`, similar to the environment in the formal semantics of first order predicate logic

For the second case, we keep an equivalence relation (implemented using a `UnionFind`) for the `VarId`s, which must first be checked before using a var. Handling variables explicitly makes it possible and convenient to store type constraints directly on vars, which can be checked on unification with a concrete type. When a new equality between two vars is discovered, their two equivalence classes are merged, and so are their constraints.

We are careful to keep these cases separate (e.g. don't substitute a variable with another variable).


src/typechecker/typesystem.cpp
```diff
@@ -253,8 +254,8 @@ bool TypeSystemCore::ll_occurs(int v, int i){
 
 int TypeSystemCore::ll_find(int i) {
 	if (ll_node_header[i].tag == Tag::Term) return i;
-	if (ll_node_header[i].data_idx == i) return i;
-	return ll_node_header[i].data_idx = ll_find(ll_node_header[i].data_idx);
+	if (m_substitution[ll_node_header[i].data_idx] == -1) return i;
+	return ll_find(m_substitution[ll_node_header[i].data_idx]);
```
SebastianMestre: This call to ll_find is unnecessary. Once a variable is substituted, it cannot change.

src/typechecker/typechecker.cpp
```diff
@@ -16,14 +16,14 @@ TypeChecker::TypeChecker(AST::Allocator& allocator)
...
+	core().new_builtin_type_function(0);  // 5  | boolean
+	core().new_builtin_type_function(0);  // 6  | unit
 
 	core().new_term(1, {}, "builtin int");    // 0 | int(<>)
 	core().new_term(2, {}, "builtin float");  // 1 | float(<>)
 	core().new_term(3, {}, "builtin string"); // 2 | string(<>)
-	core().new_term(6, {}, "builtin bool");   // 3 | boolean(<>)
-	core().new_term(7, {}, "builtin unit");   // 4 | unit(<>)
+	core().new_term(5, {}, "builtin bool");   // 3 | boolean(<>)
+	core().new_term(6, {}, "builtin unit");   // 4 | unit(<>)
```
SebastianMestre, line 26: Should use `BuiltinType` constants instead of magic numbers

src/typechecker/typesystem.cpp
```diff
@@ -157,65 +155,142 @@ void TypeSystemCore::unify_type_function(TypeFunctionId i, TypeFunctionId j) {
...
-int TypeSystemCore::compute_new_argument_count(
-    TypeFunctionData const& a_data, TypeFunctionData const& b_data) const {
 
-	if (a_data.argument_count == b_data.argument_count)
-		return a_data.argument_count;
+int TypeSystemCore::ll_find(int i) {
+	if (!ll_is_var(i)) return i;
+	VarId vi = get_var_id(i);
+	if (m_substitution[static_cast<int>(vi)] == -1) return i;
+	return ll_find(m_substitution[static_cast<int>(vi)]);
```
SebastianMestre, line 271: This call to ll_find is redundant

lushoBarlett (approved):
I would appreciate if names like `new_dummy_for_typecheck1` were replaced for their functionality. Also, the `ll_` naming scheme should be changed for a cleaner solution (like a separate object with a use-specific interface).

Besides that you have already commented on some other important aspects, I agree with those.

SebastianMestre:
> I would appreciate if names like `new_dummy_for_typecheck1` were replaced for their functionality. Also, the `ll_` naming scheme should be changed for a cleaner solution (like a separate object with a use-specific interface).

Would 100% love some more in-depth suggestions here. Things like: what would be some better names? What fields and methods could be extracted? And so on

I am honestly at a bit of a loss, so it would be neat if we could have chat about it, virtual or otherwise.

lushoBarlett:
> I am honestly at a bit of a loss, so it would be neat if we could have chat about it, virtual or otherwise.

It would be nice if we could pair-program a refactoring of this. My suggestion is that we merge this for now, but if you'd like to solve these issues right away then that's fine as well.

SebastianMestre:
> > I am honestly at a bit of a loss, so it would be neat if we could have chat about it, virtual or otherwise.
> 
> It would be nice if we could pair-program a refactoring of this. My suggestion is that we merge this for now, but if you'd like to solve these issues right away then that's fine as well.

sounds good

PR #311, opened 2023-01-11
`TRY` macro for the parser and other improvements

SebastianMestre:
In [SerenityOS/serenity](https://github.com/SerenityOS/serenity), they do error handling using a type called `AK::Result<T>` (essentially Haskell's `Either`). To avoid writing tons of code that just checks and returns the error, they came up with a really cool `TRY` macro:

```c++
auto x = TRY(failable_operation());

// essentially, expands to:
auto temp_ = failable_operation();
if (temp.is_error()) return temp.error();
auto x = temp_.value();
```

I took that and adapted it for our parser. I also made `TRY_WITH` that, instead of just returning the error, will append it to an existing error object and return that instead. (much like our old `CHECK_AND_RETURN` macro). For extra convenience, I made `REQUIRE` and `REQUIRE_WITH`, which are just calls to `Parser::require` wrapped in the other macros.

To be clear, the advantage over `CHECK_AND_RETURN` and `CHECK_AND_EXTRACT` is that the new macros behave like expressions (so they can appear within other expressions), while the old ones were full-on statements (so they'd need to be on their own line).

Using these, error handling in the parser has become much much cleaner.

----------------

I also did a few more readability changes:

- Simplify usage of `make_writer` where possible. (At some point I made it so that `Writer<SubClass*>` can be implicitly converted to `Writer<SuperClass*>`, but we never took advantage of that in the parser).
- Add `add_sub_error` method to `ErrorReport`, which makes it compatible with the new error handling macros.
- Replace `Writer<T>` with `ErrorReport` where possible.
- Add some missing error handling on `parse_expression`.

Finally, I used `TRY` instead of `TRY_WITH` in a bunch of places where we were using `CHECK_AND_RETURN`, (meaning some errors won't get wrapped with extra information about the parser's call stack) which made error messages a lot shorter yet -- i think -- a lot more readable.

-----------------

Serenity's original TRY macro can be seen here:
https://github.com/SerenityOS/serenity/blob/eed63e817422ef6582607df93d29577607202bde/AK/Try.h

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -20,27 +20,6 @@ Writer<T> make_writer(T x) {
 	return {{}, std::move(x)};
 }
 
-// WHY DO I HAVE TO TYPE THIS TWICE!?
-template <typename T, typename U>
-bool handle_error(Writer<T>& lhs, Writer<U>& rhs) {
-	if (not rhs.ok()) {
-		// NOTE: it's kinda bad that we move out of a non r-value, but
-		// due to the way we use it, it's safe.
-		lhs.add_sub_error(std::move(rhs).error());
-		return true;
-	}
-	return false;
-}
-
-template <typename T, typename U>
-bool handle_error(Writer<T>& lhs, Writer<U>&& rhs) {
-	if (not rhs.ok()) {
-		lhs.add_sub_error(std::move(rhs).error());
-		return true;
-	}
-	return false;
-}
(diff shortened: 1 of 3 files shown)


lushoBarlett (approved):
Very nice abstraction. I'm very much for it and the way it reduces LOCs. Approved.

PR #313, opened 2023-01-14
Parser initialize nodes on allocation

SebastianMestre:
I extended CSTAllocator so that it can take constructor parameters for the objects it creates. I also made all CST nodes take their members through the constructor.

This makes it so that instead of writing this code:

```c++
auto e = m_cst_allocator.make<T>();
e->m_foo = foo;
e->m_bar = bar;
e->m_baz = baz;
return make_writer(e);
```

We can write it like this:

```c++
return make_writer(m_cst_allocator.make<T>(foo, bar, baz));
```

> No need to look into it, but:
>
> - The technique used to implement `CSTAllocator::make` is called "perfect forwarding". [I wrote a blog about it a while a ago (in spanish)](https://sebmestre.blogspot.com/2022/09/sobre-perfect-forwarding-en-c.html)
> - To make it work with arbitrary amounts of parameters, I also used [template parameter packs](https://en.cppreference.com/w/cpp/language/parameter_pack).

Of course, this gets a bit long, so I added a helper on `struct Parser` to help with that:

```c++
return make_writer(make<T>(foo, bar, baz));
```

It might be a good idea to also abstract the call to `make_writer`, not sure... It doesn't feel like allocating an object and wrapping it in a Writer<T> is a concept by itself. (but to be fair, the standard library does have `std::make_unique`, `std::vector::emplace` and others)

-------------

Other that that, I noticed this pattern a lot:

```c++
auto tok = peek();
if (tok->m_type == xxx) {
    ...
} else if (tok->m_type == yyy) {
    advance_token_cursor();
    ...
} // and so on
```

Which I'm pretty sure is a piece of good 'ol premature optimization introduced by yours truly. (`match` and `consume` both call `peek` and, in this case, they'd be peeking on the same token, so we save some peek operations by checking explicitly)

I changed all those to the more readable form:

```c++
if (match(xxx)) {
    ...
} else if (consume(yyy)) {
    ...
} // and so on
```

### DIFF
--- a/src/cst.hpp
+++ b/src/cst.hpp
@@ -58,8 +58,9 @@ struct PlainDeclaration : public Declaration {
 		return identifier();
 	}
 
-	PlainDeclaration()
-	    : Declaration {CSTTag::PlainDeclaration} {}
+	PlainDeclaration(DeclarationData data)
+	    : Declaration {CSTTag::PlainDeclaration}
+	    , m_data {std::move(data)} {}
 };
 
 struct FuncDeclaration : public Declaration {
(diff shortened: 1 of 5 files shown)


lushoBarlett (approved):
Very logical improvements. 100 less lines in the parser is always a good thing. Approved.

PR #315, opened 2023-01-18
Typesafe cst

SebastianMestre:
Though it's not written anywhere in the code, there were two types of CST nodes: statements and expressions. Roughly, statements go inside seq-blocks, and expressions go on the right hand side of declarations.

I reified this separation through some subclasses of `struct CST`. It was remarkably easy to do, as the code already didn't mix these two types of nodes, except in a few cases that were simple enough to work around.

Once this was in place, some APIs could be made more type safe by distinguishing between the two, and I was able to remove some long standing runtime checks in the interpreter.

Some future steps:
- Do the same thing for `AST` (it's partially done, with `AST::Expr`. Need to add `AST::Stmt` and make APIs stronger-typed)
- Remove all references to `struct CST`, excluding those in `CSTAllocator`
- Remove tag from `struct CST`, put it in `struct Stmt` and `struct Expr`
- Separate `CSTTag` into two tag types, perhaps `CST::StmtTag` and `CST::ExprTag`

### DIFF
--- a/src/convert_ast.cpp
+++ b/src/convert_ast.cpp
@@ -469,21 +469,6 @@ static AST* convert_stmt(CST::CST* cst, Allocator& alloc) {
 #undef DISPATCH
 }
 
-AST* convert_ast(CST::CST* cst, Allocator& alloc) {
-#define DISPATCH(type)                                                         \
-	case CSTTag::type:                                                         \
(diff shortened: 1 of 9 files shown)


lushoBarlett (approved):
Looks like a good change, nothing too crazy, straight to the point. Approved.

PR #316, opened 2023-02-22
Remove token array

SebastianMestre:
TokenArray (later generalized to ChunkedArray) was created because we needed a container with address stability across push_back operations. This was needed because the lexer ran concurrently with the parser, so new tokens were created as we held and stored pointers to older tokens in the CST.

We now (for quite a while, really) fully tokenize source files before parsing. This means we don't need address stability on tokens, so we can just use a vector.

------------

While looking at this I also noticed that tokens had a really inefficient memory layout. They measured 24 bytes, and 8 of those (33%) were padding! By just reordering the members, I got rid of that padding, bringing them down to 16 bytes.

------------

The two changes combined made running the playground (without printing) on a 1MB file go from taking 0.072s to 0.063s (more than a 10% improvement). Neither the lexer nor the parser are bottlenecks in Jasper, but this is performance for absolutely free, so I'm gladly taking it.

### DIFF
--- a/src/interpreter/execute.cpp
+++ b/src/interpreter/execute.cpp
@@ -11,7 +11,6 @@
 #include "../parser.hpp"
 #include "../symbol_resolution.hpp"
 #include "../symbol_table.hpp"
-#include "../token_array.hpp"
 #include "../typechecker/ct_eval.hpp"
 #include "../typechecker/metacheck.hpp"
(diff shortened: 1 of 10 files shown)


lushoBarlett (approved):
About the bottleneck, I'm glad the file processing part is quick anyway. It's a good sign.

PR #317, opened 2023-04-02
Strongly typed ast

SebastianMestre:
Tryna give AST the same treatment that was given to CST on #315.


lushoBarlett:
Will review this weekend. Already set a reminder!

lushoBarlett (approved):
I like having Program also be different. Great change.

PR #318, opened 2023-07-09
Typechecker: Implement the typing environment in a simpler way

SebastianMestre:
In [HM systems](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system) with [the value restriction](https://en.wikipedia.org/wiki/Value_restriction), free type variables of non-values (i.e. expressions that might evaluate to mutable objects at runtime) get "bound to the environment". This is not an explicit operation. Rather, it means that those (type-level) variables just don't get generalized and will therefore be free in the typing environment (if they were generalized they wouldn't be free, but rather bound to a quantifier.)

"The typing environment" is just a fancy name for "a list of all the variables that are available in the current scope, along with their types". Since we already store types in declarations, this means we just need some data structure that allows us to store declarations and remove them according to Jasper's scoping rules.

Of course, we already have such a data structure: it's called SymbolTable. So we can use that instead of the weirdly specific CompileTimeEnvironment.

The old way of explicitly binding type variables to the environment was born out of a misunderstanding of how HM systems work. While it might be less efficient, the new way is much simpler, and probably more easily optimizable if needed.

----------------------

This change also removes the dubious "hidden variable" mechanism, which always seemed wrong and was a point of confusion for me (even though I came up with it lol). Since that mechanism doesn't actually make any logical sense as far as I can tell, there should be plenty of simple examples where it fails (i.e either over- or under-generalizes) and the new thing works correctly. It would be nice to find one to add as a regression test.

Since all this code is ridiculous, I would love to hear some potential refactorings :).

Edit: also I might've missed some calls to `declare` in `typecheck.cpp`. I will have a further look tomorrow

------------------------

Edit: This change might enable some simplifications in the type system core. I seem to remember that, when unifying typevars, we always make the newer one point to the older one. The only reason for that is how the CompileTimeEnvironment works (or used to work at some point).

Edit: Something else that became apparent with this change is that typechecking is (at least) O(N^2) on the number of declarations. This is because after processing a declaration we try to generalize its type. In order to do this we iterate over the environment (which contains all the previous declarations). It might not be easy to optimize that. This was already the case with the old code, it was just more obfuscated.


lushoBarlett:
I'm traveling in a couple hours, so I'll give a review over the weekend. Cheers.

lushoBarlett (approved):
I will have to revisit the whole code some other time to tell if more changes can be made. But removing a whole duplicated data structure and removing hidden variables is a lot in it's own right.

PR #319, opened 2023-10-20
Bytecode interpreter

SebastianMestre:
Repeatedly walking a tree is not the most efficient way to evaluate a piece of code.

There is a cool observation here: The most important thing that eval() does is fiddle with the interpreter stack and garbage collector. In fact, most of its behavior on a given run can be characterized by the sequence of method calls that it performs on those objects.

Therefore, what we could do is -- roughly -- store the sequence of method calls in an array, then read them off and perform them one by one. This will be faster when the sequence needs to be repeated multiple times because it's more cache friendly (although our ASTs are decent in that regard) and it needs to take fewer branches.

---------------

The general approach is that each Jasper function will get compiled to a list of bytecode instructions on its first call. If it succeeds, we execute bytecode. If it fails, we fall back on the old tree-walking interpreter.

This allows for the call stack to interweave bytecode functions and tree-walk functions (i.e. BC functions can call TW functions and vice-versa). This means the bytecode compiler and interpreters can be written incrementally, along several pull requests.

Since different bytecode instructions will have different sizes, we just memcpy them into an array of bytes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,7 @@ COMMON_TARGETS := \
 INTERPRETER_DIR := interpreter
 INTERPRETER_ENTRY := main
 INTERPRETER_TARGETS := \
+	bytecode \
 	error \
 	eval \
 	execute \
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -4,6 +4,7 @@
 #include "./log/log.hpp"
 #include "cst.hpp"
 #include "ast.hpp"
+#include "interpreter/bytecode.hpp"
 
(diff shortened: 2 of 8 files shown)


lushoBarlett (approved):
The idea behind this is good. The way it is implemented is straight forward and so the naming of some things was difficult for me to keep track of. An example for me was a `compile` function which calls a `visit` function which calls a specific `compile_thing` function, where `visit` is just a compilation dispatch. I got the point tho, so approved.

PR #320, opened 2023-12-01
Stack api improvements

SebastianMestre:
I encapsulated Stack a bit better, and made its usages more stack-like: no more reading the stack pointer and accessing the stack by index.

### DIFF
--- a/src/interpreter/bytecode.cpp
+++ b/src/interpreter/bytecode.cpp
@@ -95,12 +95,12 @@ static int decode(char const* stream, Interpreter::Interpreter& e) {
 
 		auto callee = value_of(e.m_stack.access(argument_count));
 
-		e.m_stack.start_stack_frame(e.m_stack.m_stack_ptr - argument_count);
+		e.m_stack.start_frame(argument_count);
 
 		eval_call_callable(callee, argument_count, e);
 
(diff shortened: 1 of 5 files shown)


lushoBarlett (approved):
Very clean. I'm wondering if the `for_each` is zero cost abstraction. I never liked that in, say, PHP. Python does that better because it lets you overload the `for-in` statement. It would be much better to be able to just write `for (auto p : m_stack)`... oh well.

SebastianMestre:
> Very clean. I'm wondering if the `for_each` is zero cost abstraction.

Yup, I measured but forgot to put the results in the PR... oops

> ... It would be much better to be able to just write `for (auto p : m_stack)`... oh well.

Oh overloading range-based for is possible by exposing .begin() and .end() methods that return iterators. 

Iterators are objects that implement a few operators:

- `++` preincrement to advance the iterator
- `!=` to compare iterators 
- `*` dereference to access the current element

This is because range for in C++ is just syntax sugar

```c++
for (auto p : m_stack) ...

// is roughly equivalent expands to

for (auto l = m_stack.begin(), r = m_stack.end(); l != r; ++l) {
  auto p = *l;
  ...
}
```

Cool idea. I will give that a shot in a separate PR

PR #321, opened 2023-12-02
Metacheck again

SebastianMestre:
I rewrote metacheck... again... for the 500th time. Good news: it no longer uses unification, it's just a top-down walk over the AST. Bad news: it rejects more programs than before. Good news: the programs we lost are pretty damn evil-looking.

So, yeah.

I think this is very close to a definitive version of meta type checking. It's way simpler and smaller (note the -178 diff), more robust, and it does its job well. It's also easy to improve accuracy and error messages because it pretty much does bidirectional type checking (just add some ifs in `check(Expr*, MetaType)`).


lushoBarlett (approved):
Very interesting observation.

PR #324, opened 2023-12-30
Remove typefunc vars

SebastianMestre:
This PR gets rid of the concept of typefunc vars and typefunc unification.

---------------------

Typefunc unification served two purposes: allow for cycles to exist between typefuncs (by first creating a variable for each typefunc that other typefuncs can refer to, then unifying each typefunc with its variable) and arity checking on type applications.

The first is addressed more directly by constructing typefuncs in two phases, as in this PR.

The second only needs to be done through unification because we can't check the arity of typefunc vars until they are unified. This is a circular justification (we need unification to deal with unification variables) so, if there is no other reason, we can get rid of it.

------------------------

The new approach:

`stub_type_function` creates the required typefunc objects, but with a typevar instead of the correct type for each field. Then, after all typefuncs are stubbed, we call `complete_type_function` on each typefunc, which actually computes the type of each field.

This is useful because references to any typefunc during `complete_type_func` can be fully resolved and checked (worst case we get a reference to a stub, but it's ok because stubs look a whole lot like the real thing).

I believe this is simpler, more elegant and more understandable than the unification approach.

------------------------

Some things I want to tackle in near-future PRs

- expose a better typevar constraint API, replacing the `new_dummy_*` family that's left over from when we heavily relied on typefunc unification to impose type constraints.
- check typevar constraints -- right now `satisfies` just returns `true`
- expose better APIs for creating type objects -- `new_term`, `ll_new_var`, `new_poly` are too annoying to read and write, so I'm thinking something more like `app`, `fresh_var`, `forall`, etc
- use `VarId` instead of `MonoId` in a bunch of places -- notably in `PolyData` and `gather_free_vars`

I actually wrote a PR that tackled a bunch of these but it was too big, intertwined and probably too annoying to review :yum:

### DIFF
--- a/src/typechecker/ct_eval.cpp
+++ b/src/typechecker/ct_eval.cpp
@@ -31,6 +31,7 @@ static AST::Constructor* constructor_from_ast(AST::Expr* ast, TypeChecker& tc, A
 static MonoId compute_mono(AST::Expr*, TypeChecker&);
 static TypeFunctionId compute_type_func(AST::Expr*, TypeChecker&);
 
+static std::vector<MonoId> compute_monos(std::vector<AST::Expr*> const&, TypeChecker&);
 
 // literals
 
@@ -433,9 +434,6 @@ static AST::Expr* ct_eval(AST::Expr* ast, TypeChecker& tc, AST::Allocator& alloc
 #undef RETURN
 }
 
-static std::unordered_map<InternedString, MonoId> build_map(
-    std::vector<InternedString> const&, std::vector<AST::Expr*> const&, TypeChecker&);
-
 static TypeFunctionId compute_type_func(AST::Identifier* ast, TypeChecker& tc) {
 	assert(ast->m_declaration);
 	assert(ast->m_meta_type == MetaType::TypeFunction);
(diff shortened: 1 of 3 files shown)


src/typechecker/ct_eval.cpp
```diff
@@ -269,27 +269,77 @@ static TypeFunctionId get_type_function_id(AST::Expr* ast) {
...
 		auto struct_expression = static_cast<AST::StructExpression*>(ast);
-		struct_expression->m_value = tc.core().new_type_function_var();
+		TypeFunctionId tf = struct_expression->m_value;
+
+		auto const& names = struct_expression->m_fields;
+		auto types = compute_monos(struct_expression->m_types, tc);
+
+		int n = names.size();
+		for (int i = 0; i < n; ++i) {
+			tc.core().ll_unify(types[i], tc.core().get_type_function_data(tf).structure[names[i]]);
```
lushoBarlett, line 283: Real chunky boi

lushoBarlett (approved):
I'm wondering if other "legacy" implementations can be simplified like this.

src/typechecker/ct_eval.cpp
SebastianMestre: Oof yeah, I just wanted to get it over with

I'll try to split it up a bit.

SebastianMestre: I extracted some variables for now as a stopgap. Maybe a better API can be added in the future.

PR #325, opened 2023-12-31
Use VarId where appropriate

SebastianMestre:
Use `VarId` instead of `MonoId` where appropriate (`PolyData` and `gather_free_vars`, mainly)

I had to define comparison operators to make it work with `std::set` and `std::map`

Also, I renamed `new_poly` to `forall`

### DIFF
--- a/src/typechecker/typecheck.cpp
+++ b/src/typechecker/typecheck.cpp
@@ -97,14 +97,14 @@ private:
 // quantifies all free variables in the given monotype
 PolyId TypecheckHelper::generalize(MonoId mono) {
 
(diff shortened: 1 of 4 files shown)


lushoBarlett (approved):
I'm wondering why's the change from `unordered_*` structures to regular ones.

SebastianMestre:
> I'm wondering why's the change from `unordered_*` structures to regular ones.

It was to avoid specializing std::hash because it's kind of ugly... I don't have a proper reason lmao

Unordered data structures are probably more appropriate here tbh

I'll roll that back

PR #328, opened 2024-01-02
Dont rewrite nodes in ct_eval

SebastianMestre:
The only remaining instances of AST rewriting were rewriting access expressions into constructor nodes and non-term identifiers into whatever they referenced.

Instead, we now compute them on-demand. Instead of rewriting access expressions into constructors when we find them, we now store the constructor in the ConstructorExpression that actually uses it.

This means that **there is no AST rewriting in ct_eval**, a sentence I've wanted to say for a long time but getting there seemed really hard lol.

I hope this change enables further simplifications.

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -233,9 +233,13 @@ struct MatchExpression : public Expr {
 	    : Expr {ExprTag::MatchExpression} {}
 };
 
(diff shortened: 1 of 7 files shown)


lushoBarlett (approved):
Looks good!

PR #329, opened 2024-01-02
Type constraint api

SebastianMestre:
I created a builder-pattern-style API for adding constraints to typevars, replacing the old idea of dummies. I think this communicates the intent a lot better when reading the code.

Also took the opportunity to reorganize TypeSystemCore's public interface... side note, I feel like it's becoming pretty cohesive thanks to the last few commits :D.

This PR actually adds more code than it removes D: (but still has negative delta due to some shenanigans :D), but I feel like the improved quality justifies it

### DIFF
--- a/src/typechecker/ct_eval.cpp
+++ b/src/typechecker/ct_eval.cpp
@@ -184,16 +184,16 @@ static AST::Constructor* constructor_from_ast(
 
 		auto access = static_cast<AST::AccessExpression*>(ast);
 
-		MonoId dummy_monotype = tc.core().ll_new_var();
-		auto v = tc.core().get_var_id(dummy_monotype);
+		MonoId actual_ty = compute_mono(access->m_target, tc);
+
+		// constraint target type
+		MonoId expected_ty = tc.core().ll_new_var();
+		auto v = tc.core().get_var_id(expected_ty);
 		tc.core().add_variant_constraint(v);
 		tc.core().add_field_constraint(v, access->m_member, tc.core().ll_new_var());
+		tc.core().ll_unify(expected_ty, actual_ty);
 
-		MonoId monotype = compute_mono(access->m_target, tc);
-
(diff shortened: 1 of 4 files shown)


lushoBarlett (approved):
I feel this is the Software Engineering class talking, but if what we strive for is clarity of the procedures, perhaps a more careful separation of responsibilities according to whatever design choices we take should be used.

What I mean is, perhaps we should build classes that represent TADs and encapsulate the data APIs that you built this PR.

But again, this is just the S.E. class talking. Do whatever you think is right.

I would love to get involved myself in this project again, since you are continuing work with it. Will try to do it as soon as I free myself from my remaining obligations. Should be a little after I'm back from vacations. Perhaps we could get together and discuss the language again.

SebastianMestre:
> I feel this is the Software Engineering class talking, but if what we strive for is clarity of the procedures, perhaps a more careful separation of responsibilities according to whatever design choices we take should be used.
> 
> What I mean is, perhaps we should build classes that represent TADs and encapsulate the data APIs that you built this PR.
> 
> But again, this is just the S.E. class talking. Do whatever you think is right.
> 
> I would love to get involved myself in this project again, since you are continuing work with it. Will try to do it as soon as I free myself from my remaining obligations. Should be a little after I'm back from vacations. Perhaps we could get together and discuss the language again.

Not sure I understand. For example, it's clear to me that the typechecker should not hand out references to its internal data structures (e.g. as in `get_type_function_data`) but I assume this is not the full extent of what you imagine. 

A note worth bringing up is that typechecker data structures use the repository pattern out of necessity, to deal with object lifetimes. This is a recurring pattern when designing good APIs in C++ and other unmanaged languages :(.

Anyways, I would love to see what you come up with :^).

I'm going to Brazil for a summer course in the 20th so let's try to schedule a meet before then 👍

PR #331, opened 2024-01-10
Jan 9 & 10 refactor

SebastianMestre:
I don't really intend on merging these changes. This is more of an exploration, to see if the ideas are good and whether to keep moving in this direction. What do you think @lushoBarlett?

- Jan 9 changes: add accessors to avoid returning references to TypeFunctionData
- Jan 10 changes: switch to strongly typed typefunc and monotype IDs (named `TypeFunc` and `Type` respectively)

----------------

The first set of changes are about hiding the in-memory representation of the type checker data structures. I think hiding such difficult design decisions is what encapsulation is good for (I'm paraphrasing *Parnas, 1972, "On the Criteria To Be Used in Decomposing Systems into Modules"* here) and should allow more free and independent modification of the core data structures with respect to the typechecking procedures.

----------------

The second set of changes is a bit less substantiated, so I need more words to explain my reasoning.

**About strongly typed IDs:** I've always thought that the fact that all of `TypeSystemCore`'s methods use plain `int` as IDs is a disaster waiting to happen. The fact that we can pass a type func ID as if it were a mono ID is slightly terrifying. However, every time I've tried to use strongly typed indices the resulting code didn't read well to me. Expressions like `type_func_data[static_cast<int>(tf)]` make want to pull my hairs.

But I think this might be caused by a lack of encapsulation. Maybe I hated the code because such expressions would end up spilling throughout and even outside of `TypeSystemCore`. Hopefully this is mitigated by the other changes.

**About the names:** In Jasper's source code we have always referred to the integers that let us index into the typechecker's data structures as "IDs", which feels like we're always working with "pointers to the thing" and not "the thing itself".

Writing code in a functional language made me appreciate how nice it would be if trees and other linked data structures were simple values that can be passed around freely. So I thought to pretend that those integers are actual objects that represent types, that they're real -- but opaque -- values and the `TypeSystemCore` just lets us inspect those values.

In that sense, `Type` and `TypeFunc` are abstract data types, with their operations exposed through `TypeSystemCore`'s interface.

Of course, using the name `TypeFunc` instead of `TypeFuncId` is just a cosmetic difference, but there is power to a name, as it affects how an entity is perceived when reading the code.


lushoBarlett (commented):
I love this quite much. My personal favorite is probably removing 'ID' semantics. I still don't know how I feel about changing `typesystem` to `core`. I guess it makes some sense considering that is the most _abstract_ file, meaning closest to the actual requirements. It implements theory as defined.

I don't believe it is necessary to separate the data definitions and operations into different classes, as some OOP enthusiasts might suggest. Technically speaking, "design-space" modules are abstract, and need not map into code.

Would be nicer for me to read, sure, since I don't remember much of what the code does. But that's a line some draw in some place and others some place else. I can just read. **NOTE:** you could make an argument that the class itself is loaded because it carries multiple ADTs and some other stuff as well.

As a personal note, didn't think you knew (or even read) Parnas. I like his theory, even tho our teacher has little time to explain _why_ we should follow it in the industry to the extent that he preaches. Objectivity is hard to come by when it comes to designing software.

lushoBarlett:
> should allow more free and independent modification of the core data structures with respect to the typechecking procedures.

Something I liked hearing during class was _every design choice is an item of change_ (Direct translation to English, haven't read the correct words in English). Which basically solves a conundrum I had with my code since forever. Implementations change all the time, we should not depend on them elsewhere.

My critique is then: theory arrives at that "Everything" limit, and with regular items of change (the ones present in the requirements) too, leaving the actual software industry with very little leeway to work with and grow (basically advocating for a _Waterfall_ software development scheme). But this is an off-topic that we could discuss in person.

PR #332, opened 2024-01-26
Simplify type creation

lushoBarlett:
This is only a proposal, I was thinking about creating a custom micro-language to write things like `\/a. a`, because a class manage this may be too much for our use case... so this is just a simplification of what we already have.

Edit: further changes can be made to the idea, I was just experimenting.

### DIFF
--- a/src/typechecker/typechecker.cpp
+++ b/src/typechecker/typechecker.cpp
@@ -45,7 +45,7 @@ TypeChecker::TypeChecker(AST::Allocator& allocator)
 	};
 
 	auto fun = [&](std::vector<Type> args, Type res) {
-		return this->core().fun(args, res);
+		return this->core().fun(std::move(args), res);
 	};
 
 	declare_builtin_value("print", forall_a(a));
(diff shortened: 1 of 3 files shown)


src/typechecker/typechecker.cpp
```diff
@@ -29,113 +29,60 @@ TypeChecker::TypeChecker(AST::Allocator& allocator)
...
+	auto forall_a = [&](Type t) {
+		return this->core().forall({aid}, t);
+	};
+
+	auto mono = [&](Type t) {
+		return this->core().forall({}, t);
+	};
+
+	auto fun = [&](std::vector<Type> args, Type res) {
+		return this->core().fun(args, res);
```
SebastianMestre, line 48: Nit: `std::move(args)`

SebastianMestre (approved):
Looks nice! As an extra suggestion, `mono_int` and friends look really clunky, so a renaming is probably in order. Maybe `integer`, `number`, etc. are better names?

About adding (essentially) a DSL for building types, I think it might be too complex but I don't want to shoot down the idea. Let's come up with a simple design after these changes are merged.

PR #333, opened 2024-06-19
Assignment refactor

SebastianMestre:
This change lays the groundwork to stop using `Reference`s to implement assignments. `References` should only be necessary for local variables that are captured by a lambda (but using them for all local variables, to avoid the complexity of handling two different cases, would be reasonable), using them to store array elements is a total hack.

I implemented this by just copy-pasting the original `CallExpression` code path that assignments currently use, inlining a bunch of stuff, replacing lookups and intermediate data structures for things that we can know statically and then applying some general simplifications.

So, the goal was to make the interpreter simpler in the future, but assignment-heavy code also got faster as a side-effect. The base10 counter microbenchmark (see below) goes from 2.92s to 2.45s (16% less) and the base2 counter (see below) goes from 1.22s to 1.08s (11% less).

base10 counter (7 nested loops):

```
fn __invoke() {
	cnt := 0;
	for (i1 := 0; i1 < 10; i1 = i1 + 1) {
		// ... etc ...
		for (i7 := 0; i7 < 10; i7 = i7 + 1) {
			cnt = cnt + 1;
		}
		// ... etc ...
	}
	return cnt; // returns 10^7
};
```

base2 counter (21 nested loops):

```
fn __invoke() {
	cnt := 0;
	for (i1 := 0; i1 < 2; i1 = i1 + 1) {
		// ... etc ...
		for (i21 := 0; i21 < 2; i21 = i21 + 1) {
			cnt = cnt + 1;
		}
		// ... etc ...
	}
	return cnt; // returns 2^21
};
```

### DIFF
--- a/src/ast.hpp
+++ b/src/ast.hpp
@@ -194,6 +194,14 @@ struct CallExpression : public Expr {
 	    : Expr {ExprTag::CallExpression} {}
 };
 
+struct AssignmentExpression : public Expr {
+	Expr* m_target;
+	Expr* m_value;
+
+	AssignmentExpression()
+		: Expr {ExprTag::AssignmentExpression} {}
+};
+
 struct IndexExpression : public Expr {
 	Expr* m_callee;
 	Expr* m_index;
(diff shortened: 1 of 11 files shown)


lushoBarlett (approved):
So this just makes the assignment operator more native than what it was. Now it's a syntax element we can operate on directly. And we have more information on it so it's faster I suppose. If I understand that correctly then, good change.

PR #334, opened 2024-06-19
Remove references from arrays

SebastianMestre:
After special-casing assignment expressions, we no longer need to store array elements through a reference. The data structure and interacting with it became simpler, and that makes me happy.

To implement this, I inlined `eval(Identifier)` and `eval(IndexExpression)` into `eval(AssignmentExpression)` and touched them up by hand. I admit it is a bit ugly, but they really are doing different things.

As a bonus, this makes some things even faster. This fibonacci benchmark goes from 1.41s before #333, to 1.34s after #333, to 1.03s now.

```
fn __invoke() {
	dp := array{};
	array_append(dp, 0);
	array_append(dp, 1);
	for (i := 0; i < 3000000; i = i+1) {
		array_append(dp, dp[i] + dp[i+1]);
	}
	return dp[3000001];
};
```

The base10 counter benchmark also got slightly faster (I'm guessing due to inlining). From 2.92s to 2.45s to 2.34s now.

```
fn __invoke() {
	cnt := 0;
	for (i1 := 0; i1 < 10; i1 = i1 + 1) {
		// ... etc ...
		for (i7 := 0; i7 < 10; i7 = i7 + 1) {
			cnt = cnt + 1;
		}
		// ... etc ...
	}
	return cnt; // returns 10^7
};
```

----------------

UPDATE: after special-casing assignment, we no longer use `Reference`s anywhere else in the interpreter, so we can have `eval(Identifier)` unwrap the reference itself instead of pushing it onto the stack.

This actually ensures an invariant: there are no temporary references on the stack, ever.

And we can exploit the invariant by removing all the `value_of` calls that were sprinkled throughout the codebase after reading a value from the stack.

Note that there are no _temporary_ references on the stack, but there are still some references. They store local variables and variables captured by lambdas.

-------------------

These changes tell me that `Reference` should really be named `Variable`, and that variables should live in a separate stack from temporaries. This would probably simplify the calling convention and `compute_offsets`.

### DIFF
--- a/src/interpreter/bytecode.cpp
+++ b/src/interpreter/bytecode.cpp
@@ -86,14 +86,14 @@ static int decode(char const* stream, Interpreter::Interpreter& e) {
 	}
 	case Instruction::Tag::GetGlobal: {
 		auto op = static_cast<GetGlobal const*>(punned);
-		e.m_stack.push(Interpreter::Value{e.global_access(op->m_name)});
+		e.m_stack.push(value_of(Interpreter::Value{e.global_access(op->m_name)}));
 		return sizeof(*op);
 	}
 	case Instruction::Tag::Call: {
 		auto op = static_cast<Call const*>(punned);
 		int argument_count = op->m_argument_count;
 
-		auto callee = value_of(e.m_stack.access(argument_count));
+		auto callee = e.m_stack.access(argument_count);
 
 		e.m_stack.start_frame(argument_count);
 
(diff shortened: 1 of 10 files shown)


lushoBarlett (approved):
I don't remember what the philosophy of values were in Jasper, our values are "moved" when assigned. So this just eliminates an intermediary, but the semantics are preserved. Right?

SebastianMestre:
> I don't remember what the philosophy of values were in Jasper, our values are "moved" when assigned. So this just eliminates an intermediary, but the semantics are preserved. Right?

Much like most dynamic languages, the semantics was that integers, booleans and numbers are copied everything else is passed/assigned by pointer. (the same as e.g. python and javascript)

This change preserves that, it just changes how variable assignment works internally.

Before, identifiers and array accesses would evaluate to an explicit reference object, which the 'value_assign' function would write through.

Now identifiers and array accesses just evaluate to their actual value, and assignment is special-cased directly in the eval function to write the assigned value in the right place.

PR #336, opened 2024-06-23
Remove gc_ptr<T>

SebastianMestre:
gc_ptr<T> was a smart pointer class that prevented the GC from deleting its pointed-to object. it allows one to store GC objects in a data structure that the GC doesn't know about, affording flexibility in how GC objects are handled on the C++ side.

But it just wasn't pulling its own weight. Ensuring that all objects we want to keep alive are on the stack when the GC runs is easy enough, and having random pointers into the GC space keeping things alive is harder to reason about.

I also renamed a couple of common methods because they had ugly-ahh names and there is no reason to put up with that. Finally, I removed `Interpreter::Error` because it was unused.

-------------------

Btw, it just keeps accidentally getting faster and faster. We are now at 0.898s on the Fibonacci benchmark (it was 1.41s before #333, a 36% speedup) and 2.04s on the base10 counter (was 2.92s, 30% speedup).


lushoBarlett (approved):
Are there any places then that need the gc ran? I reckon perhaps less places should run the gc no? Now that we know when objects die (not reachable from a stack).

SebastianMestre:
> Are there any places then that need the gc ran?

There is no one specific place that needs it, but we run it every time the number of GC-allocated objects gets "too large". (`Interpreter::run_gc_if_needed`)

More specifically, we check that condition right after an object is GC-allocated. (`Intepreter::push_list` and others)

> I reckon perhaps less places should run the gc no? Now that we know when objects die (not reachable from a stack).

Roughly, the idea would be to run the GC after popping instead of after pushing, right?

The current rule is "make sure your GC objects are on the stack when a new object is GC-allocated", which is pretty easy to uphold on `eval()`.

The new rule would be "only pop objects that you're ok with being GC-collected". I'll have to try it out. Will let you know how it goes.

PR #337, opened 2024-07-01
Typechecker core refactor

SebastianMestre:
Offline we talked about a new design for the typechecker core, and how it would require a rewrite, but it appears that the main difference is not necessarily reflected deeply in the implementation, but is rather a matter of perspective.

This perspective is that types are immutable data which must be brought up to date by `apply_substitution` before being examined, which returns a new type with all substitutions applied. The substitutions are decided through unification.

This is in contrast to the previous approach, where the abstraction was that `ll_unify` replaces every instance of a variable with its substitution, and `occurs`, `free_vars`, etc. all implicitly work on the most up to date version of a type.

(plus there is the stuff about how constraints are represented, but that can probably be changed later)

This also parallels the old implementation:

As an optimization, `ll_unify` would not replace the instances of a variable, but just write the substitution on a table, and all the other functions would have to apply pending substitutions on the fly (by calling `ll_find`, which is a shallow version of `apply_substitution`). This substitution table is now brought forward as a central piece of the algorithm.

`apply_substitution` will mutate types internally, but only as an optimization. This is possible because applying substitutions to a type which already has some substitutions applied yields the same result as applying them all in one go.

------------------

I got a bit greedy and also did some work in `typecheck.cpp`

### DIFF
--- a/src/typechecker/core.cpp
+++ b/src/typechecker/core.cpp
@@ -8,6 +8,21 @@
 TypeSystemCore::TypeSystemCore() {
 }
 
+Type TypeSystemCore::apply_substitution(Type type) {
+	if (ll_is_var(type)) {
+		VarId vi = get_var_id(type);
+		Type sub = m_substitution[static_cast<int>(vi)];
+		if (sub != Type(-1)) return apply_substitution(sub);
+		return type;
+	} else {
+		TermData& term_data = ll_term_data[data(type).data_idx];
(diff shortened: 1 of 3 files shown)


lushoBarlett (approved):
This is in line with what was previously discussed offline, approved. I see no issues, partly because the implementation is a bit large and the git diff is sadly of little help in these scenarios.

PR #338, opened 2024-07-05
Don't generalize local variables

SebastianMestre:
Earlier today I read [Let Should not be Generalised, Vytiniotis et al, 2010](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tldi10-vytiniotis.pdf). This paper outlines a bunch of bugs that can crop up if you try to generalize local variables in languages with type constraints (like field constraints in Jasper).

Aparently, this can be fixed by either capturing ALL available constraints during generalization (even if the generalized variables don't occur in them) or by lots of annoying ad-hoc work. The alternative is to not generalize locals, which eliminates this problem with some loss of expressiveness, but doesn't matter too much in practice.

In the spirit of having a working type checker that runs decently fast somewhat soon, I have decided to go with the alternative. So as of now Jasper no longer has polymorphic local variables.

### DIFF
--- a/src/typechecker/typecheck.cpp
+++ b/src/typechecker/typecheck.cpp
@@ -343,7 +343,6 @@ void typecheck_stmt(AST::Declaration* ast) {
 	// put a typevar in the decl to allow recursive definitions
 	ast->m_value_type = new_var();
 	process_contents(ast);
-	generalize(ast);
 	declare(ast);
 }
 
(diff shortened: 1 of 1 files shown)


lushoBarlett (approved):
I read the abstract and motivation sections of the paper. I am convinced that removing local generalization of variables is an acceptable loss with potential gains in expressiveness in other areas at a lower cost of development.

The fact that the change is removing a line is very funny to me. Speaks wonders of how the engine is written. It is based on the math so it doesn't surprise me. Curry-Howard isomorphism strikes again!

